PR #1, opened 2019-10-11
Make ACK optional

ray-sh:
In some case, the client don't want to receive ACK message. So, It's better we could make ACK optional for server.

Run mix test passed
Finished in 0.2 seconds
6 doctests, 30 tests, 0 failures


bryanhunter:
Closing. Added support for no-ack in v0.4.0.

PR #7, opened 2020-10-01
Properly generating the Ack

Fritzelblitz:
To Let the Ack be generated correctly, no hardcoded ACK^O01


vikas15bhardwaj:
Thanks for this PR. We opened a new PR #12 to include this fix to avoid mix.lock file update issue. Closing this PR.

PR #11, opened 2021-10-11
Handle mllp end block edge cases

starbelly:
This commit resolves an issue where two edge cases were not handled:

1. A single byte is passed to the packet framer which is the last byte of a message (carriage return). When
not handling this we might simply tack is on to the buffer and continue accumulation resulting in a failure to
respond in a timely manor (i.e.,  the client is waiting on an ACK before it sends the next message).

2. Two bytes are passed and make up the mllp trailer. The effect of not handling this case is the same as above.


lib/mllp/packet_framer.ex
```diff
@@ -74,6 +76,66 @@ defmodule MLLP.PacketFramer do
...
+                  {:ok,
+                   %{
+                     state
+                     | receiver_buffer: state.receiver_buffer <> unquote(carriage_return),
+                       current_message_type: message_type_value
+                   }}
+              end
+            end
+
+            def handle_packet(
```
vikas15bhardwaj, line 115: wouldn't this case be handled by handle_packet(packet,...) function at line 139

starbelly: You are right, beautiful catch ❤️

vikas15bhardwaj (approved):
Excellent edge case found, thank you!

PR #13, opened 2021-11-10
Increase Version with CI on main merge

vikas15bhardwaj:
This PR adds bumping of version and tagging on PR merges as part of CI.
It uses a semantic versioning based on commit convention. To read more about it:

https://github.com/mathieudutour/github-tag-action
https://semver.org/


bump_version.exs
```diff
@@ -1,11 +1,61 @@
...
-  |> Version.parse!()
-  |> (fn ver -> "#{ver.major}.#{ver.minor}.#{ver.patch + 1}" end).()
-
-File.write("VERSION", new_version)
-IO.puts("Version bumped to #{new_version}")
+defmodule Git do
+  def latest_tag() do
+    "v" <> last_tag = execute(~w[tag])
+    |> String.split("\n")
+    |> List.last()
```
calvinb: I think this will give you the tag that is last alphabetically, not semantically. You know, `"10" < "9"` kind of stuff.

.github/workflows/main.yml
```diff
@@ -43,6 +43,7 @@ jobs:
         run: mix deps.get
       - name: compile
         run: | 
+          git fetch --tags
```
starbelly, line 37: Did you find this was needed here in order for the `release` job with that one action to work? Since _build and so forth is not restored in the release job, I'm not sure this is needed. Thoughts?

bump_version.exs
```diff
@@ -1,11 +1,61 @@
...
-new_version =
-  "VERSION"
-  |> File.read!()
-  |> String.trim()
-  |> Version.parse!()
-  |> (fn ver -> "#{ver.major}.#{ver.minor}.#{ver.patch + 1}" end).()
-
-File.write("VERSION", new_version)
-IO.puts("Version bumped to #{new_version}")
+defmodule Git do
```
starbelly, line 3: Two thoughts : 

1. Do you think this should be converted into a mix task? 

2. Have you  considered replacing the version number in  `mix.exs` vs having `get_latest_version()` in its stead? Rationale behind this approach would be you can check to see if the latest tag is `:gt` than the current version in thee mix file, and then bump if it is. Thoughts?

.github/workflows/main.yml
vikas15bhardwaj: Its needed to make sure at compile time `bump_version.exs` can get `git tags`. For git tag to work, we need this

bump_version.exs
starbelly:
```
git describe --tags --abbrev=0
```
^ that should always get you  the latest tag.

.github/workflows/main.yml
starbelly: In that case I think it makes sense to move this into the  `release` job, what do you think?

bump_version.exs
vikas15bhardwaj: humm! didn't realize that. I thought it always returns in order they were applied. alternatively I found this command
`git describe --tags $(git rev-list --tags --max-count=1)` or `git describe --tags --abbrev=0` to return latest tag. This is returning the last tag as we need it. What do you think?

bump_version.exs
vikas15bhardwaj: Didn't think about having it as mix task. If we replace version in `mix.exs` file, either it's a manual step or it will require commit from CI. Manual approach, I totally want to avoid as we always miss doing this. And committing from CI kind of feels weird but I am not oppose to it. It will require some more digging though.

.github/workflows/main.yml
vikas15bhardwaj: we still would need it in compilation so that we can set lib version correctly

bump_version.exs
calvinb: > git describe --tags --abbrev=0

Waterpark used to use that, but it would occasionally return the second-latest tag, so we went brute force and implemented our own sort.

bump_version.exs
calvinb: In Waterpark we used to have the version number in a file, but then when CI bumped the version, it needed to commit a new change, which would kick off a new build, and on into Loopsville.

starbelly: Derp! Bad train of thought on my part... yeah, in retrospect, I wouldn't  suggest it :) Please ignore 2. 

Edit: 

What I do suggest though is looking for the latest tag and comparing to see if the version should be bumped. Maybe my train of that isn't well on this one, but it seems like with  the current implementation the version  will  always be `patch_level + 1` vs having a current version. That said, I need to look over this again.  

Of course, something not taken into account currently is minor and major versions. 

I really like how this is done in say rebar and rebar3_hex where by the version number is only incremented when we publish to hex and interactively, but based on the latest git tag.

starbelly: Actually, the other thing I suggested results in the same problem... I guess that's where I started to go with the last sentence in the above comment. Rather, than creating tags in ci, prefer to do this as part of a publish operation. Basically, you can take what you got here and make a wrapper task around hex.publish. That is, if the end goal is around version numbers on hex. 

Maybe it's also I'm not used to incrementing the version (even via a tag) on every commit either,  so the idea is throwing me for a loop 🤔

vikas15bhardwaj: Github action step, I used, It uses commit message convention to upgrade version. See more details here https://github.com/mathieudutour/github-tag-action . We should be matching the same in lib version. Currently yes its only doing patch no increase but based on last commit message, we can update minor or major versions too

bump_version.exs
starbelly: @calvinb  yeah, FWIW, in rebar this is done in several steps, but that's for a very generic solution where by there's no control over tag schemes. I wonder how to replicate it getting the second tag vs the first?

vikas15bhardwaj:
Closing this PR, as we have for now do manual version increase. Will look into this more with hex package version etc.

PR #14, opened 2021-11-16
Smarter reconnect behaviour for MLLP.Sender

starbelly:
This PR intends to add "smarter" reconnect behavior for MLLP.Sender. Specifically, as it stands today whenever there is an error on  send or recv  we simply attempt to reconnect. One problem as a result of the current behavior is multiple  connections will be established (i.e., the current socket is  still open, but we end up opening another one in the case of a server that multiplexes). 

This PR also aims to provide better error messages to the caller as well.


lib/mllp/sender.ex
```diff
@@ -236,8 +236,7 @@ defmodule MLLP.Sender do
         message
       end
 
-    state.tcp.send(state.socket, payload)
-    |> case do
+    case state.tcp.send(state.socket, payload) do
```
starbelly, line 237: Nit fix :) Can revert if not desired. There's a few of these, but will only comment here.

lib/mllp/sender.ex
```diff
@@ -293,54 +275,78 @@ defmodule MLLP.Sender do
...
+        handle_send_error(state, reason)
     end
   end
 
-  def handle_info({:tcp_closed, _socket}, state) do
-    telemetry(:status, %{status: :disconnected, error: :tcp_closed}, state)
-    {:noreply, maintain_reconnect_timer(state)}
+  def handle_info(:connect_timeout, state) do
+    new_state =
+      case Process.info(state.socket) do
```
starbelly: This is defensive and generally goes against the erlang way ™️ , however, being defensive here might serve us well. I also need to double check that this method of checking if the socket still  exists is  reliable. Further, per  other changes, this might not be needed.

lib/mllp/sender.ex
```diff
@@ -293,54 +275,78 @@ defmodule MLLP.Sender do
...
+
+        _ ->
+          state
+      end
+
+    reply = {:error, :send_message_failure}
+    {:reply, reply, new_state}
   end
 
-  def handle_info({:tcp_error, _, reason}, state) do
```
starbelly, line 322: Siince we set `{:active, false}` this will should be called.

lib/mllp/sender.ex
```diff
@@ -293,54 +275,78 @@ defmodule MLLP.Sender do
...
-        )
-
-        new_state = maintain_reconnect_timer(state)
-        reply = {:error, :send_message_failure}
-        {:reply, reply, new_state}
+        handle_send_error(state, reason)
     end
   end
 
-  def handle_info({:tcp_closed, _socket}, state) do
```
starbelly, line 317: Since  we set `{:active, false}` this should never be called.

lib/mllp/sender.ex
```diff
@@ -293,54 +275,78 @@ defmodule MLLP.Sender do
...
-  def handle_info({:tcp_error, _, reason}, state) do
-    telemetry(:status, %{status: :disconnected, error: :tcp_error, context: reason}, state)
-    {:stop, reason, state}
+  def terminate(reason, state) do
+    Logger.error("Sender socket terminated. Reason: #{inspect(reason)} State: %{inspect state}")
+    stop_connection(state, reason, "process terminated")
   end
 
-  def handle_info(:timeout, state) do
+  @should_reconnect_errors [:closed, :enotconn, :econnreset, :enetreset, :ehostdown]
```
starbelly: Here's the cases where we should attempt to reconnect IMO, but we should go over all possible  errors a  few time. Please refer to : https://www.erlang.org/doc/man/inet.html#type-posix

lib/mllp/sender.ex
```diff
@@ -293,54 +275,78 @@ defmodule MLLP.Sender do
...
+      %{status: :disconnected, error: error, context: "send message failure"},
       state
     )
 
-    new_state =
-      %State{state | pending_reconnect: nil}
-      |> attempt_connection()
+    new_state = maintain_reconnect_timer(state)
+    reply = {:error, {:send_message_failure, error}}
+    {:reply, reply, new_state}
```
starbelly, line 328: I don't like having the reply tuple in these functions, it was  just a way to start. I also found returning the gen_tcp error is very useful. It was  hard to tell in usagee what I ended up with message failures when eyeballing the console.

starbelly:
Note: Currently tests fail and I have added no new tests yet. I figured we may want to go over the changes themselves a bit first.

lib/mllp/sender.ex
vikas15bhardwaj: Is this more preferred way of writing case? I don't have strong opinion on either approach.

vikas15bhardwaj:
Few observation as I did some digging into how gen_tcp behaves related to connections:

- Every time we call `gen_tcp.connect` we get a new socket (which is of type `port`) if connection is successful. 
- Socket is basically a Port and port always remain open unless we call `gen_tcp.close` even if other side has closed the connection.
- On every reconnect attempt, we are basically leaving old port open currently (if any was opened previously successfully) 
- On a port, we can check Port.info to find if port is still open or not.
- We should be closing previous port before attempting to open a new one.

lib/mllp/sender.ex
starbelly: As always when it comes to style and "readability" it depends on who are you talking to. I do think this reads better. Perhaps if we were piping a few times it might be worth it, but  when ever I see a single pipe, I see that as an opportunity  to take it out. 

Piping to blocks (in this case, a case expression) is another story, and very bias on that part since  I wrote this https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/block_pipe.ex.

There's a credo check for single pipe  usage that's enabled by default iirc.  

All that said, at the end of the day it's what the team prefers. There's never a right or wrong answer, and terms like  "readability"  are impossible to define. 

I adjusted the above because my I kept having trouble reading it,  and wasn't planning on pushing it up, but it  here is none the less :)

starbelly:
> Few observation as I did some digging into how gen_tcp behaves related to connections:
> 
>     * Every time we call `gen_tcp.connect` we get a new socket (which is of type `port`) if connection is successful.
> 
>     * Socket is basically a Port and port always remain open unless we call `gen_tcp.close` even if other side has closed the connection.
> 
>     * On every reconnect attempt, we are basically leaving old port open currently (if any was opened previously successfully)
> 
>     * On a port, we can check Port.info to find if port is still open or not.


Yup, but I think I'm going to do an about face on checking the socket with `Process.info`,  I think that may be too defensive, what I have now and will push up in a few works. Basically, if we end up in that handle_info/2 callback it should be because it's correct that we end up  there, thus  avoiding the defensive check. 

I do agree, if we have an existing socket still, we should close it first (or attempt to), but let me fiddle a bit.

lib/mllp/sender.ex
vikas15bhardwaj: In my testing, I could get `closed`, `enotconn` and `econnrefused` errors. Should we add `econnrefused` to list?

lib/mllp/sender.ex
vikas15bhardwaj: We should definitely return underlying error as it helps in diagnosing issues or may be log the underlying error.

starbelly:
@vikas15bhardwaj  I have put this in a ready to review status. I think there's quite a few other changes we should make to improve reconnect error, but this solves regressions observed in integration  testing. 

I think other tickets should be opened to : 

1. Make auto-reconnect optional 

2. Add backoff for reconnect  (reconnecting over and over without backoff can be harmful) 

Those two things at the very least. But as stated, what's in this PR resolves a problem and is probably best limited to that.

.github/workflows/main.yml
```diff
@@ -11,7 +11,7 @@ env:
   MIX_ENV: test
   otp-version: '24'
   elixir-version: '1.12'
-  cache-version: '0'
+  cache-version: '1'
```
starbelly: I needed to bust the cache per an  "Old PLT found" error. If we  keep seeing this, we'll need to adjust caching of dialyzer PLTs. First time I've encountered this one.

lib/mllp/sender.ex
```diff
@@ -1,10 +1,8 @@
 defmodule MLLP.SenderContract do
-  @type sender_error ::
-          :unspecified_error
```
starbelly, line 3: These atoms were not used,  so I removed them, but can put them back off if there's a reason they were there that I'm not seeing.

lib/mllp/sender.ex
```diff
@@ -343,6 +314,48 @@ defmodule MLLP.Sender do
...
+      state
+    )
+
+    new_state = maintain_reconnect_timer(state)
+    reply = {:error, %{type: :send_failure, reason: error}}
+    {:reply, reply, new_state}
+  end
+
+  defp handle_send_error(state, error) do
+    # Add telemetry call here
```
starbelly: What do you think the  status for a telemetry call here should be? Before it was disconnected, but that's not going to be the case in any  case that hits this  clause.

lib/mllp/sender.ex
```diff
@@ -343,6 +314,48 @@ defmodule MLLP.Sender do
...
+      %{status: :disconnected, error: error, context: "receive ACK failure"},
+      state
+    )
+
+    maintain_reconnect_timer(state)
+    {:error, %{type: :recv_failure, reason: error}}
+  end
+
+  defp handle_recv_error(_state, error) do
+    # Add telemetry call here
```
starbelly: Ditto. What do you think the  status for a telemetry call here should be? Before it was disconnected, but that's not going to be the case in any  case that hits this clause.

lib/mllp/sender.ex
```diff
@@ -1,10 +1,8 @@
...
-  @type sender_error ::
-          :unspecified_error
-          | :reply_timeout_exceeded
-          | :not_connected
-          | :closed
-          | :tcp_error
+  @type error_type :: :send_error | :recv_error
+  @type error_reason :: :closed | :timeout | :inet.posix()
+
+  @type sender_error :: :not_connected | %{type: error_type(), reason: error_reason()}
```
starbelly: We discussed changing thee shape of the returned error tuples to this.  That leaves `:not_connecteed` as a different shape. I don't have strong feelings, but maybe you think the shape should be uniform?

test/sender_and_receiver_integration_test.exs
```diff
@@ -130,4 +158,29 @@ defmodule SenderAndReceiverIntegrationTest do
...
+
+    def dispatch(:mllp_hl7, message, state) when is_binary(message) do
+      reply =
+        MLLP.Ack.get_ack_for_message(
+          message,
+          :application_accept
+        )
+        |> to_string()
+
+      Process.sleep(1500)
```
starbelly: Don't like doing this, but not sure there's a better way without making more changes. Specifically right now the  reconnect timer  is hard coded at 1000 ms. If we had an option for this, we could avoid this, but I think that should be done in another PR along with other reconnect / time options and behavior. I'm not opposed to doing it here though.

lib/mllp/sender.ex
```diff
@@ -250,15 +247,7 @@ defmodule MLLP.Sender do
...
-        telemetry(
-          :status,
-          %{status: :disconnected, error: reason, context: "send message failure"},
-          state
-        )
-
-        new_state = maintain_reconnect_timer(state)
-        reply = {:error, :send_message_failure}
-        {:reply, reply, new_state}
+        handle_send_error(state, reason)
```
starbelly, line 253: I see where all the send* handle_call clauses can be collapsed into one and avoid the `handle_send_error/2` helper and `handle_recv_error/2` helpers. I would do it in another PR I think, but also not opposed to  just doing it in this PR.

lib/mllp/sender.ex
```diff
@@ -343,6 +314,48 @@ defmodule MLLP.Sender do
     stop_connection(state, reason, "process terminated")
   end
 
+  @should_reconnect_errors [
```
starbelly: I think this is about all we can unequivocally handle as far as saying  "Ok, we definitely need to try to reconnect". `:econnrefused` is an odd one, but I suppose that can happen when sending or  receiving. I think there might be other cases, but I'm reluctant to add them unless I'm 100% sure it should result in a reconnect.

vikas15bhardwaj:
In mix test, we are getting an error logged. Can you check this and may be do capture_log to avoid this showing up in console if required
.....
09:36:22.723 [error] The DefaultPacketFramer is discarding unexpected data: 
.....
..........................................................................

lib/mllp/sender.ex
vikas15bhardwaj: yes, we should make it consistent!

lib/mllp/sender.ex
```diff
@@ -343,6 +314,48 @@ defmodule MLLP.Sender do
...
+    :closed,
+    :enotconn,
+    :econnreset,
+    :enetreset,
+    :econnaborted,
+    :econnrefused
+  ]
```
vikas15bhardwaj: On this list of error, I glanced through some errors https://www.erlang.org/docs/23/man/inet.html#error_codes and found these which looks like we should include for reconnect. what do you think?
```
enetdown - Network is down
enetreset - Network dropped connection on reset (already there)
enetunreach - Network is unreachable
ehostdown - Host is down
ehostunreach - Host is unreachable
ecomm - Communication error on send
```

lib/mllp/sender.ex
vikas15bhardwaj: `econnrefused` is fine, I have seen it most when receiving port is not open

lib/mllp/sender.ex
vikas15bhardwaj: I do see each different send is doing little different things primarily to handle receive response. I like multiple smaller functions as good as they are not duplicating code

lib/mllp/sender.ex
```diff
@@ -343,6 +314,48 @@ defmodule MLLP.Sender do
...
+  defp handle_send_error(state, error) when error in @should_reconnect_errors do
```
vikas15bhardwaj: This function we should call in `attempt_connection` function also. May be we should rename it to handle_connection_error which could occur at send time or even when we are starting the process. e.g. when I do `MLLP.Sender.start_link("localhostA", 9000)` we get error `:nxdomain` and we keep trying to connect to it.

lib/mllp/sender.ex
vikas15bhardwaj: If this ever happens and we are not retrying to connect, a caller can call send message again on failure and it will most likely fail again as socket never probably recovered from previous error. Would we leave caller in a state they can't come out from without killing this process?

vikas15bhardwaj: Should we close the socket in this situation or we let the process crash?

lib/mllp/sender.ex
vikas15bhardwaj: same as above

lib/mllp/sender.ex
```diff
@@ -391,6 +404,9 @@ defmodule MLLP.Sender do
     %State{state | pending_reconnect: ref}
   end
 
+  defp maybe_close_socket(%State{socket: nil}), do: :ok
+  defp maybe_close_socket(state), do: state.tcp.close(state.socket)
```
vikas15bhardwaj: This function is similar to `stop_connection`. I also noticed, `stop_connection` is being called in reconnect function but we never reconnect after that. May be a separate bug

test/sender_and_receiver_integration_test.exs
```diff
@@ -114,6 +114,34 @@ defmodule SenderAndReceiverIntegrationTest do
...
   end
 
+  describe "timeout behaviour" do
+    test "does not reconnect when a receive timeout occurs" do
+      port = 8153
+
+      {:ok, %{pid: _receiver_pid}} =
+        MLLP.Receiver.start(
+          port,
+          SenderAndReceiverIntegrationTest.TestSlowDispatcher
```
vikas15bhardwaj: I don't see this dispatcher anywhere? The one is there is `TestNoReplySlowDispatcher`. Do we need this dispatcher here

test/sender_and_receiver_integration_test.exs
```diff
@@ -114,6 +114,34 @@ defmodule SenderAndReceiverIntegrationTest do
...
+        (HL7.Examples.wikipedia_sample_hl7() <>
+           "ZNK|JUNK|" <> String.pad_leading("\r", 100, "Z"))
+        |> HL7.Message.new()
+        |> to_string()
+
+      {:error, _} = MLLP.Sender.send_raw_and_receive_reply(sender_pid, message, 100)
+
+      assert Process.alive?(sender_pid)
+
+      Process.sleep(1000)
```
vikas15bhardwaj: I tested this after removing process.sleep and it seems to work fine every time I test it.

test/sender_and_receiver_integration_test.exs
vikas15bhardwaj: I don't see this module being used.

lib/mllp/sender.ex
starbelly: Will adjust to return thee map with `:connect_failure` as the type.

lib/mllp/sender.ex
starbelly: Fair enough 👍

lib/mllp/sender.ex
starbelly: Those make sense to me, will add.

test/sender_and_receiver_integration_test.exs
starbelly: That's because I have a `Process.sleep/1` call in the dispatcher,  I'll remove the call here.

test/sender_and_receiver_integration_test.exs
starbelly: Good catch! Fixing up. I wonder this even worked 🤔

lib/mllp/sender.ex
starbelly: I think I said to myself, let the caller decide on this one. This  is not a case we can clearly handle, so it may be best letting the caller decide. Thoughts?

test/sender_and_receiver_integration_test.exs
starbelly: Yeah, I started re-working `stop_connection` to make use of maybe_close_socket/1, but  then there was a reason I paused and  said these are not the same cases, but I can look again and re-work it.

lib/mllp/sender.ex
starbelly: I may be wrong, but that's the  job of `maintain_reconnect_time` is  it not? It's not optimal though I agree. But, the idea as I understand it was, ensure the connection timer is active, after 1 second (or less) the process will be sent  an info message which will attempt a reconnect. 

Are you thinking we should try to reconnect once and if it doesn't pan out bail?

lib/mllp/sender.ex
```diff
@@ -236,29 +234,23 @@ defmodule MLLP.Sender do
...
-    state.tcp.send(state.socket, payload)
-    |> case do
+    case state.tcp.send(state.socket, payload) do
       :ok ->
-        reply =
-          if Keyword.get(options, :expect_reply) do
-            receive_reply(state, Keyword.get(options, :reply_timout))
-          else
-            {:ok, :sent}
+        if Keyword.get(options, :expect_reply) do
```
starbelly, line 239: I don't like this form at all, but I would not opt to refactor in this PR. Let's do a "clean up" PR instead.

test/sender_and_receiver_integration_test.exs
```diff
@@ -114,6 +115,36 @@ defmodule SenderAndReceiverIntegrationTest do
     end
   end
 
+  describe "timeout behaviour" do
+    test "does not reconnect when a receive timeout occurs" do
```
vikas15bhardwaj: Do you think, this test description now should change to "does not add additional process/leak ports on reconnect"

vikas15bhardwaj: And I think if you't use `SenderAndReceiverIntegrationTest.TestSlowDispatcher` in test, it will still work same way because message never ends well and it times out due to very low timeout.

we can also remove `TestNoReplyDispatcher`, we are not using it anywhere

starbelly: Can change the description, n/p. 

>  And I think if you't use SenderAndReceiverIntegrationTest.TestSlowDispatcher in test, it will still work same way because message never ends well and it times out due to very low timeout. we can also remove TestNoReplyDispatcher, we are not using it anywhere

Once again, good catch. That's also a bug though. That is, if I'm specifying a module that doesn't exist and it doesn't result in a crash, it's a bug. Will open an issue for this. 


If you change sender.ex back to the version from `main` and remove the `Process.sleep/1` call you'll see it passes when it shouldn't

vikas15bhardwaj: yes, It passes on main as expected but fails if we add `Process.sleep/1` which is expected because we don't close port before attempting the connection again

lib/mllp/sender.ex
vikas15bhardwaj: We decided on not keep error list for retry and will instead will try to reconnect. We will close old connection though before attempting the connection again.

vikas15bhardwaj (approved):
Thank you for working on this with me!

PR #21, opened 2021-11-24
Rename MLLP.DefaultDispatcher to MLLP.EchoDispatcher

### LINKED ISSUE #17: Rename DefaultDispatcher to EchoDispatcher
In EchoDispatcher, we should remove `Logger.error` message. We should echo the message as info and returns positive ACK. It may require some changes in README as well.

_Originally posted by @vikas15bhardwaj in https://github.com/HCA-Healthcare/elixir-mllp/pull/16#discussion_r755466773_

starbelly:
- changed MLLP.DefaultDispatcher to MLLP.EchoDispatcher
 - updated references where required
 - updateed README.md
 
 Closes #17

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ Next, start an MLLP.Sender process and store its PID.
 
 Alternatively, you could start a Sender using a DNS name rather than an IP address.
 ```
-{:ok, s1} = MLLP.Sender.start_link("localhost", port: 4090, MLLP.EchoDispatcher)
+{:ok, s1} = MLLP.Sender.start_link("localhost", 4090)
 ```
 
 
@@ -47,7 +47,7 @@ You will see log info like...
  }}
 ```
 
-The Logger debug part tells us that the Receiver received the HL7 message. The Logger error part says we haven't set a Dispatcher and the message will not be routed anywhere other than to the console log. Finally, the return value (`{:error,  :application_reject, ...}`) is a NACK. The DefaultDispatcher will not reply with an `:application_accept`.
+The Logger debug part tells us that the Receiver received the HL7 message. The Logger error part says we haven't set a Dispatcher and the message will not be routed anywhere other than to the console log. Finally, the return value (`{:error,  :application_reject, ...}`) is a NACK. The EchoDispatcher will not reply with an `:application_accept`.
 
 Now, we will stop the Receiver.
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -22,7 +22,7 @@ Next, start an MLLP.Sender process and store its PID.
 
 Alternatively, you could start a Sender using a DNS name rather than an IP address.
 ```
-{:ok, s1} = MLLP.Sender.start_link("localhost", 4090)
+{:ok, s1} = MLLP.Sender.start_link("localhost", port: 4090, MLLP.EchoDispatcher)
```
vikas15bhardwaj: We don't need dispatched on `Sender.start_link`

vikas15bhardwaj: Also the port stays as parameter on Sender

lib/mllp/echo_dispatcher.ex
```diff
@@ -1,7 +1,8 @@
-defmodule MLLP.DefaultDispatcher do
+defmodule MLLP.EchoDispatcher do
```
bryanhunter, line 1: By renaming here you will also need to update the docs (receiver.ex line 51) to use the new name.

As is `iex(1)> h MLLP.Receiver.start` tells user to do it wrong

README.md
starbelly: Good catch.

lib/mllp/echo_dispatcher.ex
starbelly: Good catch.

vikas15bhardwaj:
Couple of places, I still see DefaultDispatcher in comments. README line 50, and 104

README.md
```diff
@@ -55,7 +55,11 @@ Now, we will stop the Receiver.
  MLLP.Receiver.stop(4090)
 ```
 
-### Custom message dispatcher
+### Writing a message dispatcher
+
+MLLP does not ship with a default message dispatcher as the cases can vary signficantly from domain to domain. Instead,
```
bryanhunter:
```suggestion
MLLP does not ship with a default message dispatcher as the cases can vary significantly from domain to domain. Instead,
```

README.md
```diff
@@ -47,15 +47,19 @@ You will see log info like...
  }}
 ```
 
-The Logger debug part tells us that the Receiver received the HL7 message. The Logger error part says we haven't set a Dispatcher and the message will not be routed anywhere other than to the console log. Finally, the return value (`{:error,  :application_reject, ...}`) is a NACK. The DefaultDispatcher will not reply with an `:application_accept`.
+The Logger debug part tells us that the Receiver received the HL7 message. The Logger error part says we haven't set a Dispatcher and the message will not be routed anywhere other than to the console log. Finally, the return value (`{:error,  :application_reject, ...}`) is a NACK. The EchoDispatcher will not reply with an `:application_accept`.
```
bryanhunter:
```suggestion
The Logger `[debug]` part tells us that the Receiver received the HL7 message. The Logger `[info]` part explains that the EchoDispatcher does not route the message anywhere (just "logs and discards" messages). Finally, the return value (`{:error,  :application_reject, ...}`) is a NACK. The EchoDispatcher will not reply with an `:application_accept`.
```

vikas15bhardwaj (approved):

PR #24, opened 2021-11-24
Add additional flags to dialyzer config

starbelly:
This commit adds :unmatched_returns and :race_conditions to dialzyer config.

See https://www.erlang.org/doc/man/dialyzer.html for details

- changed lib/mllp/sender.ex to cancel timer asynchronously


starbelly:
Note : I need to squash this before merging as the original commit message needs to be altered  per switching from `async: true` to `info: false`

vikas15bhardwaj (approved):

PR #27, opened 2021-11-29
Simplify MLLP.Sender/Client api and standardize errors 

starbelly:
- reduce the various send_* functions to send/2,/3,/4 and send_async/2,3
- add auto_reconnect_interval option
- wrap and unwrap payloads by default
- only verify a response with MLLP.Ack if the payload provided is an HL7.Message.t()
- move the initial connection attempt into a handle_continue/2 clause
- normalize (validate) host parameter to a tuple and raise if invalid
- Always log an error when we fail to establish a connection
- Store connection failures in state so we can return the caller the last known error in case
   they try to send a message while the connection is still being established
- Add and standardize on MLLP.Client.Error (exception) for returned errors
- Use Config vs  Mix.Config in config/*
- misc refactors, enhancements, and fixups


lib/mllp/sender.ex
```diff
@@ -1,32 +1,24 @@
...
 
-  @callback send_hl7_and_receive_ack(
+  @type send_options :: %{
+          optional(:expect_reply) => boolean(),
+          optional(:timeout) => non_neg_integer(),
+          optional(:wrap_message) => boolean(),
+          optional(:reply_timeout) => non_neg_integer() | :infinity
+        }
+
+  @callback send(
```
vikas15bhardwaj: I see we currently have some non_hl7 and raw send functions, which I believe are not used normally. I haven't seen anyone requesting non_hl7 messages on MLLP. We do have some use case to receive non_hl7 on MLLP.

@bryanhunter - Do you see any reason we should keep non_hl7 send support?

lib/mllp/sender.ex
```diff
@@ -1,32 +1,24 @@
...
-  @callback send_non_hl7(pid :: pid, payload :: String.t()) ::
-              {:ok, :sent} | {:error, sender_error()}
-  @callback send_raw_and_receive_reply(
-              pid :: pid,
-              payload :: String.t(),
-              timeout :: non_neg_integer()
-            ) ::
-              {:ok, String.t()} | {:error, sender_error()}
-  @callback send_raw(pid :: pid, payload :: String.t()) :: {:ok, :sent} | {:error, sender_error()}
+              {:ok, binary() | String.t()}
```
vikas15bhardwaj: wouldn't we always return ACK message or error? In what cases we return `{:ok, binary()}`

starbelly: Good catch.  That's actually redundant. It should be just `{:ok, String.t()}`  `String.t()` is equivalent to `binary()`

starbelly: > wouldn't we always return ACK message or error? 

I might have misunderstood. We would return a binary (or `String.t())` if verify_reply is false (i.e., send raw and receive raw reply or send %HL7{} but verify_reply is set to false).

lib/mllp/sender.ex
```diff
@@ -76,133 +68,91 @@ defmodule MLLP.Sender do
...
 
-  @spec send_raw_and_receive_reply(
+  @default_send_opts %{
+    expect_reply: true,
+    timeout: 5000,
+    wrap_message: true,
+    reply_timeout: :infinity
+  }
+
+  @spec send(
```
starbelly: There's two things I don't like about this function name. 

1. It can be seen as redundant `Sender.send()` maybe `Client.send()` would read better
2. It sends AND receives, which is why I suspect it was `send_and_receive*` before yet, I don't like having an `and` in the function name either. OTOH, we can say, well sometimes sending implies a response and in the case of MLLP/HL7 that's always true. 

Suggestions and thoughts appreciated. What I do like about it, it's nice and short ;)

vikas15bhardwaj: I am ok with name `send`, but `Sender.send()` looks little weird and I like `Client.send()`. It's anyway is a MLLP client sending messages to a server.

lib/mllp/sender.ex
vikas15bhardwaj: I take `raw` message here still an HL7 message but as `binary()` rather than type HL7.Message.t(). I think `send` should accept both `binary` and `HL7.Message.t()`. If message received is an `HL7.Message.t()`, then we should always return verified ACK. If we receive `binary()` raw hl7, we return `binary()` response. I think, we don't need `verify_reply` option

starbelly: Ahh yeah, it does, but I failed to add this to the callback spec. Thanks!

lib/mllp/sender.ex
starbelly: Let's rename it to Client then :) 👍

lib/mllp/sender.ex
```diff
@@ -1,450 +0,0 @@
-defmodule MLLP.SenderContract do
```
starbelly, line 1: This should not been a delete since I did `git mv`. I'll have to fix this up.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
+defmodule MLLP.ClientContract do
```
starbelly, line 1: For some reason  git is  being stubborn about marking this as a new  file vs moved and  changed. I can try to fix  it, let me know.

config/config.exs
```diff
@@ -1,6 +1,6 @@
 # This file is responsible for configuring your application
 # and its dependencies with the aid of the Mix.Config module.
-use Mix.Config
+import Config
```
starbelly, line 3: Went ahead and changed this since Mix.Config is deprecated

config/test.exs
```diff
@@ -1,3 +1,3 @@
-use Mix.Config
+import Config
```
starbelly, line 1: Went ahead and changed this since Mix.Config is deprecated

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+          port :: non_neg_integer(),
+          options :: [keyword()]
+        ) :: {:ok, pid()}
+
+  def start_link(address, port, options \\ []) do
+    checked_address = normalize_address!(address)
+    GenServer.start_link(__MODULE__, [address: checked_address, port: port] ++ options)
+  end
+
+  defp normalize_address!({_, _, _, _} = addr), do: addr
```
starbelly: I think this should be moved to another module because it could be tested as a pure function, but I would opt to do that  in another PR.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+
+    opts
+    |> Map.merge(@default_opts)
+    |> Map.put_new(:tcp, socket_module)
+    |> Map.put(:pid, self())
+    |> Map.put(:tls_opts, opts.tls)
+    |> Map.put(:send_opts, send_opts)
+  end
+
+  def put_socket_address(%{address: address, port: port} = opts) do
```
starbelly: This should be `defp`, but will wait to make that  change.  This is merely so we don't have to constantly construct the socket address every time we want to log it. 

The caller can override the defaults at startup or on send/*

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+    verify_ack(raw_ack, raw_message)
+  end
+
+  defp verify_ack(raw_ack, raw_message) do
+    ack = Envelope.unwrap_message(raw_ack)
+    unwrapped_message = Envelope.unwrap_message(raw_message)
+    Ack.verify_ack_against_message(unwrapped_message, ack)
+  end
+
+  defp new_error(context, error) do
```
starbelly, line 419: This is a move to standardize on MLLP.Client.Error for what we return (or raise). `:message` is present if the caller (or  us in a future PR) wish to  raise  and provide a decent error message. 

`:type` was changed to `:context` as that's really what we're saying with `:type` previously (i.e., providing a context). 

`:reason` merely holds the raw error if the caller wishes to match on  it or do something with it.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+    %MLLP.Client.Error{
+      reason: error,
+      context: context,
+      message: format_error(error)
+    }
+  end
+
+  def format_error({:tls_alert, _} = err) do
```
starbelly: Added mostly for logging  purposes.

lib/mllp/client/error.ex
```diff
@@ -0,0 +1,5 @@
+defmodule MLLP.Client.Error do
```
starbelly, line 1: See https://github.com/HCA-Healthcare/elixir-mllp/pull/27/files#r770810192

test/client_and_receiver_integration_test.exs
```diff
@@ -1,6 +1,7 @@
-defmodule SenderAndReceiverIntegrationTest do
+defmodule ClientAndReceiverIntegrationTest do
```
starbelly, line 1: Renamed to align  with renaming Sender to Client.

test/client_and_receiver_integration_test.exs
```diff
@@ -262,39 +297,39 @@ defmodule SenderAndReceiverIntegrationTest do
     end
 
     @tag :tls
-    @tag port: 8154
+    @tag port: 8155
```
starbelly, line 291: Had to change port numbers all the way down. There's a better  way of handling this, namely using port 0 and providing the randomly chosen port to the test. This IMO should  be done in another PR.

test/client_and_receiver_integration_test.exs
```diff
@@ -396,36 +433,44 @@ defmodule SenderAndReceiverIntegrationTest do
...
 
-    @tag port: 8161
+    @tag port: 8162
     @tag client_cert: ""
     @tag keyfile: ""
-    @tag reason:
-           {:certificate_required,
-            'TLS client: In state connection received SERVER ALERT: Fatal - Certificate required\n'}
+
+    if :erlang.system_info(:otp_release) == '22' do
```
starbelly: This has to be done as per the changes around the case when  socket is `nil`, our return is more consistent, but the charlist (string) varies from OTP ver  to OTP ver (consistent between OTP 23 and 24, but diff on 22). 

I have thoughts on improving all this, but should probably be  done in another PR. But, mainly it's  usually not  advantageous to match on exact strings like this as it makes for brittle tests, but the way the tests have been DRYed up makes working  in a better way of handling this not a small change.

test/client_test.exs
```diff
@@ -0,0 +1,316 @@
+defmodule ClientTest do
```
starbelly, line 1: Git  doing the same thing here, says the file was added vs renamed / moved.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
+defmodule MLLP.ClientContract do
+  @type error_type :: :connect_failure | :send_error | :recv_error
+  @type error_reason :: :closed | :timeout | :no_socket | :inet.posix()
+
+  @type sender_error :: MLLP.Client.Error.t()
+
+  @type options :: [
```
starbelly, line 7: Some  of the names of these options could perhaps be better, but not a big deal either.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+    opts = [:binary, {:packet, 0}, {:active, false}] ++ state.socket_opts ++ state.tls_opts
+
+    case state.tcp.connect(state.address, state.port, opts, 2000) do
+      {:ok, socket} ->
+        state1 = ensure_pending_reconnect_cancelled(state)
+        telemetry(:status, %{status: :connected}, state1)
+        %{state1 | socket: socket, connect_failure: nil}
+
+      {:error, reason} ->
+        message = format_error(reason)
```
starbelly, line 335: We now log on a  failure to connect always, we also store the last connect failure error in state in the case the caller tries to send when we the connection is still trying to be established.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+    opts
+  end
+
+  @default_opts %{
+    telemetry_module: MLLP.DefaultTelemetry,
+    tls_opts: [],
+    socket_opts: []
+  }
+
+  @default_send_opts %{
```
starbelly, line 384: We optimize for the case that the caller expect MLLP.Client to  wrap the message and expects a reply, and that  reply is to be verified via MLLP.Ack.

test/client_and_receiver_integration_test.exs
```diff
@@ -164,13 +169,29 @@ defmodule SenderAndReceiverIntegrationTest do
...
 
-      assert {:error, %{type: :recv_failure, reason: :closed}} ==
-               MLLP.Sender.send_raw_and_receive_reply(sender_pid, "Simple message")
+      # We have to set up an assertion like below because we 
+      # may error out on send, but more than likely on recv.
+      # Even if we set send_timeout to 1ms it may not fail on send. It's
+      # highly depenent on the OS's gen_tcp backend, tcp/ip stack and tcp/ip
+      # settings. 
+      res =
+        case MLLP.Client.send(sender_pid, "Simple message") do
```
vikas15bhardwaj: Can we use this code instead. It will match on any context value unless matching on any context value is an issue

```
assert match?(
               {:error, %Error{context: _, message: "connection closed"}},
               MLLP.Client.send(sender_pid, "Simple message")
             )
```

test/client_and_receiver_integration_test.exs
vikas15bhardwaj: correct, I also didn't like using so many ports. I tried using only one port but it fails when parallel tests are running in CI. Locally it worked fine

starbelly: Interesting, I'll give it a second set of eyes later today, may simply be not possible in CI, but we could also just make a  custom random port function in tests ourselves (i.e., get all listening ports, pick  a  port number not being used).

test/client_and_receiver_integration_test.exs
vikas15bhardwaj: If you like, we can ignore message matching by making following changes to these tests:
```
  describe "client cert validation" do
    setup ctx do
      verify = ctx[:verify] || :verify_peer
      allowed_clients = ctx[:allowed_clients] || []

      tls_options = [
        cacertfile: "tls/root-ca/ca_certificate.pem",
        certfile: "tls/server/server_certificate.pem",
        keyfile: "tls/server/private_key.pem",
        verify: verify
      ]

      transport_opts = %{tls: tls_options}

      {:ok, %{pid: receiver_pid}} =
        MLLP.Receiver.start(
          port: ctx.port,
          dispatcher: MLLP.EchoDispatcher,
          transport_opts: transport_opts,
          allowed_clients: allowed_clients
        )

      client_cert = ctx[:client_cert] || "tls/client/client_certificate.pem"
      keyfile = ctx[:keyfile] || "tls/client/private_key.pem"

      sender_tls_options = [
        verify: :verify_peer,
        cacertfile: "tls/root-ca/ca_certificate.pem",
        certfile: client_cert,
        keyfile: keyfile
      ]

      tls_alert = ctx[:reason] || []

      expected_error_reasons = [:einval, :no_socket, :closed] ++ tls_alert

      on_exit(fn -> MLLP.Receiver.stop(ctx.port) end)

      [
        receiver_pid: receiver_pid,
        sender_tls_options: sender_tls_options,
        expected_error_reasons: expected_error_reasons
      ]
    end

    @tag port: 8161
    @tag verify: :verify_none
    @tag client_cert: ""
    @tag keyfile: ""
    test "does not verify client cert if verify none option is provided on receiver", ctx do
      make_call_and_assert_success(ctx, ctx.ack)
    end

    @tag port: 8162
    @tag client_cert: ""
    @tag keyfile: ""

    @tag reason: [:handshake_failure, :certificate_required]
    test "no peer cert", ctx do
      make_call_and_assert_failure(ctx, ctx.expected_error_reasons)
    end

    @tag port: 8163
    test "accepts a peer cert", ctx do
      make_call_and_assert_success(ctx, ctx.ack)
    end

    @tag port: 8164
    @tag allowed_clients: ["client-1"]
    test "accepts peer cert for allowed client", ctx do
      make_call_and_assert_success(ctx, ctx.ack)
    end

    @tag port: 8165
    @tag allowed_clients: ["client-x", "client-y"]
    test "reject peer cert for unexpected clients", ctx do
      expected_error_reasons = [:einval, :closed]

      log =
        capture_log(fn ->
          make_call_and_assert_failure(ctx, expected_error_reasons)
        end)

      assert log =~ ":fail_to_verify_client_cert"
    end

    @tag port: 8166
    @tag client_cert: "tls/expired_client/client_certificate.pem"
    @tag keyfile: "tls/expired_client/private_key.pem"

    @tag reason: [:certificate_expired]
    test "reject expired peer cert", ctx do
      make_call_and_assert_failure(ctx, ctx.expected_error_reasons)
    end

    @tag port: 8167
    @tag client_cert: "tls/server/server_certificate.pem"
    @tag keyfile: "tls/server/private_key.pem"

    @tag reason: [:handshake_failure]
    test "reject server cert as peer cert", ctx do
      make_call_and_assert_failure(ctx, ctx.expected_error_reasons)
    end

    @tag port: 8168
    @tag allowed_clients: ["client-x", "client-1"]
    test "accept peer cert from multiple allowed clients", ctx do
      make_call_and_assert_success(ctx, ctx.ack)
    end

    defp make_call_and_assert_success(ctx, expected_result) do
      assert expected_result == start_sender_and_send(ctx)
    end

    defp make_call_and_assert_failure(ctx, expected_error_reasons) do
      reason =
        start_sender_and_send(ctx)
        |> case do
          {:error, %Error{reason: {:tls_alert, {reason, _}}}} -> reason
          {:error, %Error{reason: reason}} -> reason
        end

      assert reason in expected_error_reasons
      # assert expected_error_reasons |> Enum.any?(&match?(^reason, &1))
    end

    defp start_sender_and_send(ctx) do
      {:ok, sender_pid} =
        MLLP.Client.start_link("localhost", ctx.port, tls: ctx.sender_tls_options)

      MLLP.Client.send(
        sender_pid,
        HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new()
      )
    end
  end
```

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
+defmodule MLLP.ClientContract do
+  @type error_type :: :connect_failure | :send_error | :recv_error
+  @type error_reason :: :closed | :timeout | :no_socket | :inet.posix()
+
+  @type sender_error :: MLLP.Client.Error.t()
```
vikas15bhardwaj: should be renamed to client_error

starbelly: Nice catch  :)

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+          address ::
+            :inet.ip4_address()
+            | :inet.ip6_address()
+            | String.t(),
+          port :: non_neg_integer(),
+          options :: [keyword()]
+        ) :: {:ok, pid()}
+
+  def start_link(address, port, options \\ []) do
+    checked_address = normalize_address!(address)
```
vikas15bhardwaj: I think, we can rename variable here to `normalized_address`

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+        # hostname
+        String.to_charlist(addr)
+    end
+  end
+
+  defp normalize_address!(addr) when is_list(addr), do: parse_address!(addr)
+
+  defp normalize_address!(addr) when is_atom(addr), do: addr
+
+  defp normalize_address!(addr), do: raise("Invalid server ip address : #{inspect(addr)}")
```
vikas15bhardwaj: Should we raise `MLLP.Client.Error` here?

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+
+  defp parse_address!(addr) do
+    case :inet.parse_address(addr) do
+      {:error, _} ->
+        raise "Invalid server ip address : #{inspect(addr)}"
```
vikas15bhardwaj: and here too?

lib/mllp/client.ex
starbelly: I think a  more appropriate error would be ArgumentError, wdyt?

lib/mllp/client.ex
starbelly: I removed the intermediate var all  together now, but can put back in place if you like.

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+        )
+
+        state
+        |> maintain_reconnect_timer()
+        |> Map.put(:connect_failure, reason)
+    end
+  end
+
+  defp maintain_reconnect_timer(state) do
+    ref = state.pending_reconnect || Process.send_after(self(), :timeout, 1000)
```
vikas15bhardwaj: Should we put this time in Config, so we can have different delays for tests and reduce sleep time in tests

lib/mllp/client.ex
vikas15bhardwaj: any reason we can't make it private now

lib/mllp/client.ex
starbelly: Yeah, makes perfect sense to me. Was going to do that in another PR, but  might as  well do it here :)

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+      context: context,
+      message: format_error(error)
+    }
+  end
+
+  def format_error({:tls_alert, _} = err) do
+    to_string(:ssl.format_error({:error, err}))
+  end
+
+  def format_error(:closed), do: "connection closed"
```
vikas15bhardwaj: Are these formatting for logging purpose. Can we not simply do `inspect` on error?

starbelly: Yes,  for logging purpose,  I figured it's friendlier (and in some cases  cheaper) to avoid an inspect/1

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+            :inet.ip4_address()
+            | :inet.ip6_address()
+            | String.t(),
+          port :: non_neg_integer(),
+          options :: [keyword()]
+        ) :: {:ok, pid()}
+
+  def start_link(address, port, options \\ []) do
+    checked_address = normalize_address!(address)
+    GenServer.start_link(__MODULE__, [address: checked_address, port: port] ++ options)
```
vikas15bhardwaj: I always wonder, why we should create multiple processes for same address and port. Does it introduce process leaks Or it limits us for processing the messages. I think leaving it up to caller is OK. What do you think?

lib/mllp/client.ex
```diff
@@ -0,0 +1,444 @@
...
+
+  @type send_options :: %{
+          optional(:expect_reply) => boolean(),
+          optional(:timeout) => non_neg_integer(),
+          optional(:wrap_message) => boolean(),
+          optional(:reply_timeout) => non_neg_integer() | :infinity,
+          optional(:verify_reply) => boolean()
+        }
+
+  @callback send(
```
vikas15bhardwaj, line 17: Couple of points on this:

- Should we have two methods `send_async` and `send`
     - send_async - as the name suggest does not wait for reply but can send reply using a callback. It also takes away `exact_reply` option
     - send - is synchronous and waiting for reply

On `verify_reply` - It's about checking reply if its valid HL7 ACK or not. Should we do it only if payload is `HL7.Message.t()`, otherwise we don't need to do it. And we can remove this flag also.

On `wrap_message` - Should we do it if message is not wrapped without using the flag?

lib/mllp/client.ex
starbelly: I'd leave it up to the caller. A lot of MLLP servers in the wild from what I've  read are multiplexing.

lib/mllp/client.ex
starbelly: > Should we have two methods `send_async` and `send`
> 
>     * send_async - as the name suggest does not wait for reply but can send reply using a callback. It also takes away `exact_reply` parameter
> 
>     * send - is synchronous and waiting for reply


I  like  that idea a lot, I very much don't like all the options that can be manipulated. The less options the better, but see below. 

> On verify_reply - It's about checking reply if its valid HL7 ACK or not. Should we do it only if payload is HL7.Message.t(), otherwise we don't need to do it. And we can remove this flag also.

That makes sense  to me. I simply didn't know all the cases we would want to support. In that regard, I think `send_async` is not even really needed per the MLLP protocol itself. Can  you think of an actual case where a caller would want to sent and not receive a reply? I  can not. 

> On wrap_message - Should we do it if message is not wrapped without using the flag?

I thought about that, that makes perfect sense to me,  but it comes back to the same question.  Are  there actually any cases where someone would send data that's not enveloped? I think a lof  of the decisions I made around options was merely to support the existing test cases, but I didn't feel great about them.

vikas15bhardwaj: `wrap_message` -  Most caller shouldn't worry about wrapping message at their end. It's a MLLP protocol thing and I think it should be handled here.
`send_async` - I haven't seen this case either, but since we have its implementation, we can keep it. It may be useful in cases where someone wants to just blast off messages without worrying about ACKs during some sort of test.

starbelly: Should  we always  unwrap per getting rid of wrap_message option? Makes sense to me.

lib/mllp/client.ex
starbelly: Yup, we  can

vikas15bhardwaj:
Some minor comments:
- Can we add `send_async` to `ClientContract`
- We should change `:sender` to client in one telemetry event on line 366 in client module
- In multiple tests, we are using `sender` as variable names for client, should we change those?

vikas15bhardwaj (approved):

PR #28, opened 2021-11-29
Remove MLLP.Receiver.start/0

starbelly:
- Removed start/0 since options are mandatory now

Closes #28


vikas15bhardwaj (approved):
Every drop counts!

PR #30, opened 2021-12-01
Downgraded ranch version to 1.8.0

vikas15bhardwaj:
Downgrading `ranch` dependency to 1.8. Some app fails to run due to dependency on 1.8 version due to `cowboy`. 

Testing instructions:
Follow README section Using TLS


starbelly (approved):
approved

PR #32, opened 2021-12-07
Added client ip restrictions

### LINKED ISSUE #31: Add optionally Client IP restrictions
If a listener is provided a list of IPs, then only client running on these IPs should be allowed. We would be using this on wristband app to restrict access.

vikas15bhardwaj:
#31 

Adding client ip restrictions to restrict client from specific list of IPs. When we start listener, we can optionally provide list of IPs from where client connections should be allowed.

Testing instructions:
- By default, as before it will allow all clients
- Start Receiver `MLLP.Receiver.start([port: 9000, dispatcher: MLLP.EchoDispatcher, allowed_client_ips: ["127.0.0.0"]])`
- Start Sender `MLLP.Sender.start_link("localhost", 9000)`
     A warning should appear [warn]  Unexpected client {{127, 0, 0, 1}, 59223} is attempting to connect
- If try to send message using sender pid, MLLP.Sender.send_hl7_and_receive_ack(sender_pid, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new()), will get error `{:error, %{reason: :closed, type: :recv_failure}}`

### DIFF
--- a/lib/mllp/receiver.ex
+++ b/lib/mllp/receiver.ex
@@ -365,7 +365,7 @@ defmodule MLLP.Receiver do
     |> String.to_charlist()
     |> :inet.gethostbyname()
     |> case do
-      {:ok, {:hostent, _, _, _, _, [address]}} ->
+      {:ok, {:hostent, _, _, _, _, [address | _]}} ->
         address
 
       error ->
--- a/lib/mllp/receiver.ex
+++ b/lib/mllp/receiver.ex
@@ -350,14 +350,27 @@ defmodule MLLP.Receiver do
   defp normalize_ip(ip) when is_binary(ip) do
     ip
     |> String.to_charlist()
-    |> :inet.getaddr(:inet)
+    |> :inet.parse_address()
     |> case do
       {:ok, address} ->
         address
 
+      _ ->
+        normalize_hostname(ip)
+    end
+  end
+
+  defp normalize_hostname(name) do
+    name
+    |> String.to_charlist()
+    |> :inet.gethostbyname()
+    |> case do
+      {:ok, {:hostent, _, _, _, _, [address]}} ->
+        address
+
       error ->
         Logger.warn(
-          "IP #{inspect(ip)} provided is not a valid IP #{inspect(error)}, filtering this IP"
+          "IP/hostname #{inspect(name)} provided is not a valid IP/hostname #{inspect(error)}. It will be filtered from allowed_client_ips list"
         )
 
         nil
--- a/lib/mllp/receiver.ex
+++ b/lib/mllp/receiver.ex
@@ -344,10 +344,11 @@ defmodule MLLP.Receiver do
     end
   end
 
-  defp normalize_ip({_, _, _, _} = ip), do: ip
-  defp normalize_ip({_, _, _, _, _, _, _, _} = ip), do: ip
+  def normalize_ip({_, _, _, _} = ip), do: ip
+  def normalize_ip({_, _, _, _, _, _, _, _} = ip), do: ip
+  def normalize_ip(ip) when is_atom(ip), do: normalize_ip(to_string(ip))
 
-  defp normalize_ip(ip) when is_binary(ip) do
+  def normalize_ip(ip) when is_binary(ip) do
     ip
     |> String.to_charlist()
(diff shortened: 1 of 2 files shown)


lib/mllp/receiver.ex
```diff
@@ -328,4 +343,30 @@ defmodule MLLP.Receiver do
       framing_context2
     end
   end
+
+  defp normalize_ip({_, _, _, _} = ip), do: ip
```
starbelly: One suggestion :

I would suggest the latest version of this function I had in this WIP : https://github.com/HCA-Healthcare/elixir-mllp/pull/27/files#diff-4fc1cf0a1c259a35f3a58eb89534250fbbad4a234f6845390f94d916220a6343R79

I'm not sure about getaddr/2 call below. That'll perform a dns lookup  and in the case the ip is an ipv4 or ipv6 as a binary it will fail  with `{:error, :einval}` 


To validate an ip address in the case that it's a binary is tricky. You can run `:inet.parse*` but if it's  a host name it'll fail. 

You could try one if it fails, then try the other, if both fail we know it's one of two things : 

1. The ip address is bad, possibly the hostname is bad
2. The hostname might be fine,but name resolution failed. 

Is your intent that name resolution must succeed anyway?

vikas15bhardwaj: Glad you raised this point. I glanced through methods of `inet` and two close ones were `getaddr/2` and `parse_address` for parsing binary IP address. 
I noticed, `getaddr/2` could also parse 'localhost', '127.0.0.1' and {127, 0, 0, 1} but `parse_address` could not as it only worked with binary. But then `getaddr` also takes `address_family` as parameter which is basically if it's IPV4 vs IPV6. I chose IPV4 only because I felt there are not many client with MLLP protocol use IPV6 (I may be wrong).

Both `parse_address` and `getaddr` don't do dns lookup, whatever DNS I tried it failed except for `localhost`. And I thought we really don't need to use DNS and just use IP. 
 
After thinking through this more though, this is what we should do:

- use `parse_address` for binary IP
- if parse_address fails, try use `gethostbyname/1` (but it also returns a list of IPs, not sure why and in which case. DNS I tried, I got one IP in list)
- if both fails, then its invalid IP or DNS

thoughts?

starbelly: Ahh yeah, you are right, it will work with an ip address. The folly of going back and forth betwen elixir and erlang is I'll see `"1.1.1.1"` and sometimes my brain says "binary!" and then sometimes it says "string" (char list) :p 

Yeah, I think your proposal is sound so long as dns failure should trigger the error so to speak. Since this will only happen with and  `allowed_ips` option and not effect the general case, it should be fine :) We might find ourselves coming  back to make an option for this for this option later, but that's fine too :) Kaizen!

starbelly: Ahh but not, it does indeed do  dnslookup ... 

```
:inet.getaddr('erlef.org', :inet)
{:ok, {66, 42, 87, 202}}
``` 

```
nslookup erlef.org
...

Non-authoritative answer:
Name:	erlef.org
Address: 66.42.87.202
```

We can see this in action here : https://github.com/erlang/otp/blob/23c95823c669ae071c6bc30b30dd7225fb4bad8d/lib/kernel/src/inet.erl#L1398

vikas15bhardwaj: Interesting, yup `getaddr` does work for dns lookup. I may have missed something when trying. It looks like it is using `gethostname` method internally as well. In order to add IPV6 support though we would be better off using `parse_address/1` and `gethostbyname/1` method as suggested above. It will cover both IPV4 and IPV6.

If DNS lookup, or parse address fails, we can filter out failed address from the list of allowed IPs with a warning Or raise an error. I am currently filtering with a warning

starbelly: Since the option is `allowed_ips` how about we simply raise (or warn) if it's not an  ip address vs trying to do a DNS lookup?

vikas15bhardwaj: I can change `allowed_client_ips` to `allowed_clients`

starbelly: That'll work too 👍

starbelly (approved):

PR #34, opened 2021-12-14
Added client cert validation

vikas15bhardwaj:
This PR adds support for validating clients using certificates #19 and completes TLS cycle. 

Key points:
- In order for listeners to enforce clients to send a certificate, start listener with transport_opts having `tls` options including `verify: :verify_peer`
- listener which uses `:ranch.handshake` (which internally use `:ssl.handshake`) performs ssl handshake with client.
- listener also validates client cert Issued to ensure it trusts the cert issued to before accepting connection

Follow instructions on README to go through client cert validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7(
 ```
 
 ## Using Client Certificates
-MLLP listener can force client to provide a valid certificate before establishing a successful connections. Follow steps below to use client certs with listener
+MLLP listener can enforce client to provide a valid certificate before establishing a successful connection. Follow steps below to use client cert with a listener
 
 ### Start MLLP listener with :verify_peer option
 ```
@@ -228,7 +228,7 @@ iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7(
 MLLP listener supports two options to restrict incoming client connections to make sure it accepts only trusted clients.
 
 1) IP/DNS restriction - In this mode, we can restrict incoming connections using Client IP/DNS.
-2) Client Cert Check - If `verify: :verify_peer` option is enabled on listener, it will enforce client to send a valid client cert and will only allow connection if it trusts the certificate issued to.
+2) Client Cert Check - If `verify: :verify_peer` option is enabled on listener, it will enforce client to send a valid client cert and will only allow the connection if certificate returned from the client is valid and trusted.
 
 Here are couple of exmaples of using client restrictions
 ### Options 1 - Client IP/DNS restrictions
@@ -249,7 +249,7 @@ iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7(
 
 `[warn]  Failed to verify client {ip, port}, error: :client_ip_not_allowed`
 
-***This example provided is without TLS, We can modify it to use with TLS. Make sure to specify `verify: :verify_none` option in transport_opts of listener. See [Using TLS](#using-tls): for details on setting up TLS connections.***
+***This example provided is without TLS, We can modify it to use with TLS. Make sure to specify `verify: :verify_none` option in transport_opts on the listener. See [Using TLS](#using-tls): for details on setting up TLS connections.***
 
 ### Options 2 - Client Cert Check
 #### Start MLLP listener with TLS and allowed_clients options
@@ -265,7 +265,7 @@ iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify
 ```
 iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new())
 ```
-***In such scenarios starting a sender with a valid certificate but if cert issued to is not one of the trusted client by listener, connection will fail and a warning will be logged on the listener***
+***In the above scenarios we start a sender with a valid certificate, but the cert issued is not one of the trusted client by the listener, thus the connection fails and a warning is logged by the listener***
 
 `[warn]  Failed to verify client {ip, port}, error: :fail_to_verify_client_cert`
 
--- a/README.md
+++ b/README.md
@@ -185,14 +185,15 @@ The first step in TLS configuration is to create a TLS certificates, which can b
 
 `sh tls/tls.sh`
 
-This script create the following certs:
+This script creates the following certs:
 - root ca
 - server certificate signed by root ca
 - client certificate signed by root ca
+- expired client certificate signed by root ca
 ### Start Receiver
 
 ```
-iex> MLLP.Receiver.start(port: 8154, dispatcher: MLLP.EchoDispatcher, transport_opts: %{tls: [cacertfile: "tls/root-ca/ca_certificate.pem", verify: :verify_peer, certfile: "tls/server/server_certificate.pem", keyfile: "tls/server/private_key.pem"]})
+iex> MLLP.Receiver.start(port: 8154, dispatcher: MLLP.EchoDispatcher, transport_opts: %{tls: [cacertfile: "tls/root-ca/ca_certificate.pem", verify: :verify_none, certfile: "tls/server/server_certificate.pem", keyfile: "tls/server/private_key.pem"]})
 ```
 
 ### Start Sender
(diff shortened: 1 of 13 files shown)


lib/mllp/receiver.ex
```diff
@@ -192,17 +193,16 @@ defmodule MLLP.Receiver do
...
-    allowed_clients =
-      Keyword.get(opts, :allowed_clients, [])
-      |> Enum.map(&normalize_ip/1)
-      |> Enum.reject(&is_nil(&1))
+    verify_peer = get_in(transport_opts, [:socket_opts, :verify]) == :verify_peer
 
+    allowed_clients = get_allowed_clients(verify_peer, opts)
     proto_mod = __MODULE__
 
     proto_opts = [
```
starbelly: This could be map (proto opts is declared as `any()` in ranch). Then we could just work with a map internally for beautiful and efficient pattern matching 😍  Thoughts?

lib/mllp/peer.ex
```diff
@@ -0,0 +1,80 @@
...
+          :client_info => {:inet.ip_address(), :inet.port_number()}
+        }
+
+  @type error_type ::
+          :client_ip_not_allowed
+          | :fail_to_verify_client_cert
+
+  @spec validate(t(), Keyword.t()) :: {:ok, :success} | {:error, error_type()}
+
+  def validate(peer, options) do
```
starbelly: If we went with thee suggestion here : https://github.com/HCA-Healthcare/elixir-mllp/pull/34/files#r769075119 
then we could have one validate/2 function that merely matches on the options.This would allow us to avoid passing true/false around  (boolean blindness),  wdyt?

lib/mllp/peer.ex
```diff
@@ -0,0 +1,80 @@
...
+    allowed_clients = Keyword.get(options, :allowed_clients, [])
```
starbelly: If we transformed allowed_clients into a map, then we could use `is_map_key/2 guard in the new Peer module  and avoid a conditional `if` and `else` in the body. What do you think? 



Edit:

Sorry, to be  clear transform the  list into a map around here : https://github.com/HCA-Healthcare/elixir-mllp/pull/34/files#diff-1ab91908935d40d578aaf4788d1b1ce5b45115c2072a319cb0e40a9c5b59e625R204

lib/mllp/peer.ex
```diff
@@ -0,0 +1,80 @@
...
+    presented == reference
+  end
+
+  defp fqdn_fun({:cn, value}) do
+    value
+  end
+
+  defp validate_client_ip(_, []), do: {:ok, :success}
+
+  defp validate_client_ip({ip, _port}, allowed_clients) do
```
starbelly: if we went with this suggestion : https://github.com/HCA-Healthcare/elixir-mllp/pull/34/files#r769077503 

Then here we  could have : 

```
defp validate_client_ip({ip, _port}, allowed_clients) when is_map_key(ip, allowed_clients) do 
  {:ok, :success} 
 end
 
defp validate_client_ip(_, _), do: {:error, :client_ip_not_allowed}

File: lib/mllp/receiver.ex [hunk 5]
```diff
@@ -344,6 +362,17 @@ defmodule MLLP.Receiver do
     end
   end
 
+  defp get_allowed_clients(true, opts) do
+    Keyword.get(opts, :allowed_clients, [])
+    |> Enum.map(&to_charlist/1)
```
starbelly, line 367: Along with the other suggestions, we could transform this into a map : 

```
opts 
|>  Keyword.get(:allowed_clients, [])
|> Enum.into(%{}, fn(ip) -> {ip, true}) end)
```

lib/mllp/peer.ex
```diff
@@ -0,0 +1,80 @@
...
+         fqdn_fun: &fqdn_fun/1,
+         match_fun: &match_fun/2
+       ) do
+      {:ok, :success}
+    else
+      {:error, :fail_to_verify_client_cert}
+    end
+  end
+
+  defp match_fun({:cn, reference}, {_, presented}) do
```
starbelly: Here's a fun trick and it should apply here : 

```
  defp match_fun({:cn, reference}, {_, reference}), do: true 
  
  defp match_fun(reference, {_, reference}), do: true
  
  defp match_fun(_, _), do: false 
```

Thoughts?

README.md
```diff
@@ -205,6 +206,69 @@ iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify
 iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new())
 ```
 
+## Using Client Certificates
+MLLP listener can force client to provide a valid certificate before establishing a successful connections. Follow steps below to use client certs with listener
```
starbelly:
```suggestion
MLLP listener can force a client to provide a valid certificate before establishing a successful connection. Follow the steps below to use client certs with a listener.
```

README.md
```diff
@@ -205,6 +206,69 @@ iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify
...
+### Send a message
+```
+iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new())
+```
+
+***In this example starting a sender on another server other than localhost will fail and a warning will be logged on the server***
+
+`[warn]  Failed to verify client {ip, port}, error: :client_ip_not_allowed`
+
+***This example provided is without TLS, We can modify it to use with TLS. Make sure to specify `verify: :verify_none` option in transport_opts of listener. See [Using TLS](#using-tls): for details on setting up TLS connections.***
```
starbelly:
```suggestion
***This example provided is without TLS, We can modify it to use with TLS. Make sure to specify `verify: :verify_none` option in transport_opts on the listener. See [Using TLS](#using-tls): for details on setting up TLS connections.***
```

README.md
```diff
@@ -205,6 +206,69 @@ iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify
...
+### Send a message
+```
+iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new())
+```
+
+## Using Client Restrictions
+MLLP listener supports two options to restrict incoming client connections to make sure it accepts only trusted clients.
+
+1) IP/DNS restriction - In this mode, we can restrict incoming connections using Client IP/DNS.
+2) Client Cert Check - If `verify: :verify_peer` option is enabled on listener, it will enforce client to send a valid client cert and will only allow connection if it trusts the certificate issued to.
```
starbelly:
```suggestion
2) Client Cert Check - If `verify: :verify_peer` option is enabled on the listener, it will enforce the client to send a valid client cert and will only allow the connection if certificate returned from the client is valid and trusted.
```

README.md
```diff
@@ -205,6 +206,69 @@ iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify
...
+### Start MLLP Sender
+```
+iex> {:ok, s3} = MLLP.Sender.start_link("localhost", 8154, tls: [verify: :verify_peer, cacertfile: "tls/root-ca/ca_certificate.pem", certfile: "tls/client/client_certificate.pem", keyfile: "tls/client/private_key.pem"])
+```
+
+### Send a message
+```
+iex> MLLP.Sender.send_hl7_and_receive_ack(s3, HL7.Examples.wikipedia_sample_hl7() |> HL7.Message.new())
+```
+***In such scenarios starting a sender with a valid certificate but if cert issued to is not one of the trusted client by listener, connection will fail and a warning will be logged on the listener***
```
starbelly:
```suggestion
***In the  above scenario we start a sender with valid certificate, but the cert issued  is not one of the trusted client by the listener, thus the connection fails and a warning is logged by the listener***
```

mix.lock
```diff
@@ -18,6 +18,7 @@
   "mix_test_watch": {:hex, :mix_test_watch, "1.0.2", "34900184cbbbc6b6ed616ed3a8ea9b791f9fd2088419352a6d3200525637f785", [:mix], [{:file_system, "~> 0.2.1 or ~> 0.3", [hex: :file_system, repo: "hexpm", optional: false]}], "hexpm", "47ac558d8b06f684773972c6d04fcc15590abdb97aeb7666da19fcbfdc441a07"},
   "mox": {:hex, :mox, "1.0.0", "4b3c7005173f47ff30641ba044eb0fe67287743eec9bd9545e37f3002b0a9f8b", [:mix], [], "hexpm", "201b0a20b7abdaaab083e9cf97884950f8a30a1350a1da403b3145e213c6f4df"},
   "nimble_parsec": {:hex, :nimble_parsec, "1.2.0", "b44d75e2a6542dcb6acf5d71c32c74ca88960421b6874777f79153bbbbd7dccc", [:mix], [], "hexpm", "52b2871a7515a5ac49b00f214e4165a40724cf99798d8e4a65e4fd64ebd002c1"},
+  "ok": {:hex, :ok, "1.11.2", "0d7a72ec1450598623d2c559fca01e2fafaa7c62b31769273fd59a610b44cca8", [:mix], [], "hexpm", "330125d5c9deb116f92b35bb188aaeba7857e93205bc27a7604b279f4971c432"},
```
starbelly: Need to `mix deps.unlock && git  add mix.lock && git commit && git push` to remove this.

starbelly (approved):

starbelly:
beautiful

PR #38, opened 2022-01-03
Adjust echo dispatcher to return an ack or nack depending on the validity of a message

starbelly:
This modifies the echo dispatcher to respond with an ack or nack depending on the message being deemed  as  valid or vs  always returning a nack. 

It also modifies the function head to accept any type where before it would crash if it was given `:mllp_unknown` as the  type.

### DIFF
--- a/lib/mllp/echo_dispatcher.ex
+++ b/lib/mllp/echo_dispatcher.ex
@@ -9,21 +9,32 @@ defmodule MLLP.EchoDispatcher do
 
   @behaviour MLLP.Dispatcher
 
-  @spec dispatch(:mllp_hl7, binary(), MLLP.FramingContext.t()) :: {:ok, MLLP.FramingContext.t()}
-  def dispatch(:mllp_hl7, message, state) when is_binary(message) do
+  @spec dispatch(:mllp_hl7 | :mllp_unknown, binary(), MLLP.FramingContext.t()) ::
+          {:ok, MLLP.FramingContext.t()}
+  def dispatch(type, message, state) when is_binary(message) do
     Logger.info(
-      "The EchoDispatcher simply logs and discards messages. Message type: mllp_hl7 Message: #{message}"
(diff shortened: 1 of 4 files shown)


vikas15bhardwaj:
It looks good, just one small thing. If we send a raw message e.g. "TEST" it returns with following
```
{:ok and "MSH|^~\\&|????|????|????|????|2022-01-13 23:09:05.402811Z||ACK^O01|-576460752303423483|P|2.5\rMSA|AR|-576460752303423483|Message was not parsable as HL7. Reason: missing_header_or_encoding\r"}
```
  

May be we can just return "mllp_unknown" back in response

starbelly:
> It looks good, just one small thing. If we send a raw message e.g. "TEST" it returns with following
> 
> ```
> {:ok and "MSH|^~\\&|????|????|????|????|2022-01-13 23:09:05.402811Z||ACK^O01|-576460752303423483|P|2.5\rMSA|AR|-576460752303423483|Message was not parsable as HL7. Reason: missing_header_or_encoding\r"}
> ```
> 
> May be we can just return "mllp_unknown" back in response


That makes sense  to me 👍

lib/mllp/echo_dispatcher.ex
```diff
@@ -11,9 +11,13 @@ defmodule MLLP.EchoDispatcher do
 
   @spec dispatch(:mllp_hl7 | :mllp_unknown, binary(), MLLP.FramingContext.t()) ::
           {:ok, MLLP.FramingContext.t()}
-  def dispatch(type, message, state) when is_binary(message) do
+  def dispatch(:mllp_unknown, _, state) do
+    {:ok, %{state | reply_buffer: "mllp_unknown"}}
```
vikas15bhardwaj: We should wrap reply message `MLLP.Envelope.wrap_message()`. It generates an additional debug message.

starbelly: Done.

vikas15bhardwaj (approved):

PR #39, opened 2022-01-03
Add module docs for MLLP.Client

starbelly:
Screen Shot 2022-01-03 at 5 33 12 PM

### DIFF
--- a/lib/mllp/client.ex
+++ b/lib/mllp/client.ex
@@ -13,7 +13,7 @@ defmodule MLLP.ClientContract do
           reply_timeout: non_neg_integer() | :infinity,
           socket_opts: [:gen_tcp.option()],
           telemetry_module: nil,
-          tls_opts: [:ssl.tls_client_option()]
+          tls: [:ssl.tls_client_option()]
         ]
 
   @type send_options :: %{
@@ -176,7 +176,7 @@ defmodule MLLP.Client do
 
   ## Options
 
-  * `:auto_reconnect_interval' - Specify the interval between connection attempts. Specifically, if an attempt
+  * `:auto_reconnect_interval` - Specify the interval between connection attempts. Specifically, if an attempt
      to establish a connection fails, either post-init or at some point during the life span of the client, the value
      of this option shall determine how often to retry a reconnection. Defaults to 1000 milliseconds.
 
@@ -187,14 +187,14 @@ defmodule MLLP.Client do
      Note that `:binary`, `:packet`, and `:active` can not be overriden.
 
   * `:telemetry_module` - A callback module for which MLLP.Client will call `YourTelmetryModule.exectute/3`. The 
-    callback will be passed an event name, meaasurements, and metadata. 
+    callback will be passed an event name, measurements, and metadata. 
     Defaults to `MLLP.DefaultTelemetry`.
 
-  * `:tls_opts` - A list of tls options as supported by [`:ssl`](`:ssl`). When using TLS it is highly recommended you
-     set `:verify` to `:verify_peer`, select a CA trust store using the `:cacertfile` or `:cacerts` options. Addtional 
-     hardening may be acheived though other ssl options  such as enabling certificate revocation via the 
-     `:crl_check` and `:crl_cache` options and customization of enabled protocols and cipher suites for your 
-     specific use-case. See [`:ssl`](`:ssl`) for details.
+  * `:tls` - A list of tls options as supported by [`:ssl`](`:ssl`). When using TLS it is highly recommended you
+     set `:verify` to `:verify_peer`, select a CA trust store using the `:cacertfile` or `:cacerts` options. 
+     Addtionally further hardening may be acheived though other ssl options such as enabling 
+     certificate revocation via the `:crl_check` and `:crl_cache` options and customization of 
+     enabled protocols and cipher suites for your specific use-case. See [`:ssl`](`:ssl`) for details.
 
   """
   @spec start_link(
--- a/lib/mllp/client.ex
+++ b/lib/mllp/client.ex
@@ -5,9 +5,10 @@ defmodule MLLP.ClientContract do
   @type client_error :: MLLP.Client.Error.t()
 
   @type options :: [
+          auto_reconnect_interval: non_neg_integer(),
           reply_timeout: non_neg_integer() | :infinity,
-          tls_opts: map(),
-          socket_opts: map()
+          socket_opts: [:gen_tcp.option()],
+          tls_opts: [:ssl.tls_client_option()]
         ]
 
   @type send_options :: %{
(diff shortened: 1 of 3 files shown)


lib/mllp/client.ex
```diff
@@ -5,9 +5,11 @@ defmodule MLLP.ClientContract do
   @type client_error :: MLLP.Client.Error.t()
 
   @type options :: [
+          auto_reconnect_interval: non_neg_integer(),
           reply_timeout: non_neg_integer() | :infinity,
-          tls_opts: map(),
-          socket_opts: map()
+          socket_opts: [:gen_tcp.option()],
```
starbelly, line 14: This was not correct.

lib/mllp/client.ex
```diff
@@ -104,13 +164,36 @@ defmodule MLLP.Client do
...
+
+  ## Options
+
+  * `:reply_timeout` - Optionally specify a timeout value for receiving a response. Must be a positive integer or 
+     `:infinity`. Defaults to `:infinity`.
+    
+  * `:socket_opts` -  A list of socket options as supported by [`:gen_tcp`](`:gen_tcp`). 
+     Note that `:binary`, `:packet`, and `:active` can not be overriden.
+
+  * `:telemetry_module` - A callback module for which MLLP.Client will call `YourTelmetryModule.exectute/3`. The
```
starbelly: Puposefully  did  not document this in  detail as the current telemetry parameters and usage overall needs polish and revision. As  an example,  for metadata we pass internal state to the callback in whole

lib/mllp/client.ex
```diff
@@ -42,12 +106,7 @@ defmodule MLLP.Client do
   @behaviour ClientContract
 
   @type pid_ref :: atom | pid | {atom, any} | {:via, atom, any}
-  @type ip_address ::
```
starbelly, line 45: This was not needed as `:inet.socket_address() | String.t()`  covers it all.

test/client_test.exs
```diff
@@ -2,7 +2,6 @@ defmodule ClientTest do
   use ExUnit.Case, async: false
   alias MLLP.Client
   alias MLLP.Client.Error
-  doctest Client
```
starbelly, line 5: I'm not sure why this was ever doc tested, but it's not something that can effectively be doc tested (i.e., it's  impure).

lib/mllp/client.ex
```diff
@@ -34,6 +41,67 @@ defmodule MLLP.ClientContract do
...
+  @moduledoc """
+  MLLP.Client provides a simple tcp client for sending and receiving data i
+  via [MLLP](https://www.hl7.org/documentcenter/public/wg/inm/mllp_transport_specification.PDF) over TCP.
+
+  While MLLP is primarily used to send [HL7](https://en.wikipedia.org/wiki/Health_Level_7) messages, 
+  MLLP.Client can be used to send non-hl7 messages, such as XML.
+
+  ## Connection Behaviour 
+
+  Upon successful start up via `start_link/4`, the  client will attempt to establish a connection to the given address
```
starbelly, line 52: There's actually a bit more to document here but I feel some changes in the future that might effect what would have been documented  (specifically, when we deem a  connection to be closed).

lib/mllp/client.ex
```diff
@@ -34,6 +40,67 @@ defmodule MLLP.ClientContract do
 end
 
 defmodule MLLP.Client do
+  @moduledoc """
+  MLLP.Client provides a simple tcp client for sending and receiving data i
```
vikas15bhardwaj: Should this be just sending data? also there is letter "i" doing something alone

lib/mllp/client.ex
```diff
@@ -34,6 +40,67 @@ defmodule MLLP.ClientContract do
...
+  on the provided port. If a connection can not be immediately established, the client will keep 
+  trying to establish a connection per the value of `:auto_reconnect_interval` which defaults to 
+  1 second. Therefor it is possible that before a connection is fully established, the caller
+  may attempt to send a message which will result in `MLLP.Client.Error.t()` being returned containing
+  the last error encountered in trying to establish a connection. Additionally, said behavour could be encountered
+  at any point during life span of an MLLP.Client process if the connection becomees severed on either side. 
+
+  All connection, send, and receive failures will be logged as errors.
+
+  ## Examples
```
vikas15bhardwaj, line 62: I think we should start a receiver so examples can work or point it to `MLLP.Receiver` module to start receiver.

lib/mllp/client.ex
```diff
@@ -104,13 +167,40 @@ defmodule MLLP.Client do
...
+  ## Options
+
+  * `:auto_reconnect_interval` - Specify the interval between connection attempts. Specifically, if an attempt
+     to establish a connection fails, either post-init or at some point during the life span of the client, the value
+     of this option shall determine how often to retry a reconnection. Defaults to 1000 milliseconds.
+
+  * `:reply_timeout` - Optionally specify a timeout value for receiving a response. Must be a positive integer or 
+     `:infinity`. Defaults to `:infinity`.
+    
+  * `:socket_opts` -  A list of socket options as supported by [`:gen_tcp`](`:gen_tcp`).
```
vikas15bhardwaj, line 207: double `[:gen_tcp]`

lib/mllp/client.ex
vikas15bhardwaj: I think, we don't need to include telemetry_module if we just emit some meaningful telemetry events using telemetry.span or execute.

lib/mllp/client.ex
```diff
@@ -104,13 +167,40 @@ defmodule MLLP.Client do
...
+     `:infinity`. Defaults to `:infinity`.
+    
+  * `:socket_opts` -  A list of socket options as supported by [`:gen_tcp`](`:gen_tcp`). 
+     Note that `:binary`, `:packet`, and `:active` can not be overridden.
+
+  * `:telemetry_module` - A callback module for which MLLP.Client will call `YourTelmetryModule.exectute/3`. The 
+    callback will be passed an event name, measurements, and metadata. 
+    Defaults to `MLLP.DefaultTelemetry`.
+
+  * `:tls` - A list of tls options as supported by [`:ssl`](`:ssl`). When using TLS it is highly recommended you
```
vikas15bhardwaj, line 210: double :ssl

lib/mllp/client.ex
```diff
@@ -104,13 +167,40 @@ defmodule MLLP.Client do
...
+     set `:verify` to `:verify_peer`, select a CA trust store using the `:cacertfile` or `:cacerts` options. 
+     Additionally, further hardening can be achieved though other ssl options such as enabling
```
vikas15bhardwaj: is it "through" other ssl

lib/mllp/client.ex
```diff
@@ -120,12 +210,35 @@ defmodule MLLP.Client do
...
   def is_connected?(pid), do: GenServer.call(pid, :is_connected)
 
+  @doc """
+  Instructs the client to disconnect (if connected) and attempt a reconnect.
+  """
   @spec reconnect(pid :: pid()) :: :ok
   def reconnect(pid), do: GenServer.call(pid, :reconnect)
 
+  @doc """
+  Sends a message and and receives a response.
```
vikas15bhardwaj: double "and"

lib/mllp/client.ex
```diff
@@ -170,6 +289,12 @@ defmodule MLLP.Client do
     GenServer.call(pid, {:send_async, payload}, timeout)
   end
 
+  @doc """
+  Stops an MLLP.Client given a MLLP.Client pid.
+
+  This function will always return `:ok` per `GenServer.stop/1`, thus
+  you may give it a pid tha references a client that is already stopped.
```
vikas15bhardwaj: missing "t" from "that"

lib/mllp/client.ex
starbelly: I agree with this. I was simply documented what was there. I will undocument this and open an issue or PR to remove the option.

lib/mllp/client.ex
starbelly: vim derp :)

lib/mllp/client.ex
starbelly: Ahh, this is actually just link i.e., `:gen_tcp` ends up pointing to the `:gen_tcp` module document.

lib/mllp/client.ex
starbelly: Ditto. Just ends up becoming a link to ssl docs in erlang/otp.

lib/mllp/client.ex
starbelly: Done.

vikas15bhardwaj (approved):

PR #41, opened 2022-01-13
Add context attribute to MLLP.Receiver and pass to FramingContext

starbelly:


vikas15bhardwaj:
Wondering what do you think of renaming `custom_data` to `context`

starbelly:
> Wondering what do you think of renaming `custom_data` to `context`

That sounds good to me. Do you want to adjust the name on framing context as well?

vikas15bhardwaj:
let's rename it to `context` on receive side and in FramingContext we can add it as `receiver_context` to be explicit

vikas15bhardwaj (approved):

Issue #18, opened 2021-11-24
MLLP.Sender reconnection backoff and options


starbelly:
Currently, for MLLP.Sender auto re-connection happens at a fixed interval. This could be harmful to the servers you're connecting to. Instead, there should be exponential  backoff and perhaps an option to give up after  N attempts, perhaps crash. 

Likewise, there should be options to configure backoff behavior. 

There should probably also be an option to opt out of auto reconnect all together.

starbelly:
Closing as resolved via #48

PR #48, opened 2022-05-23
Add backoff to client connection handling

pggalaviz:
Addresses issue #18  

Adds an exponential backoff to handle reconnection of client. This is an option available that defaults to false.

### DIFF
--- a/lib/mllp/client.ex
+++ b/lib/mllp/client.ex
@@ -11,6 +11,7 @@ defmodule MLLP.ClientContract do
   @type options :: [
           auto_reconnect_interval: non_neg_integer(),
           use_backoff: boolean(),
+          backoff_max_seconds: integer(),
           reply_timeout: non_neg_integer() | :infinity,
           socket_opts: [:gen_tcp.option()],
           telemetry_module: nil,
@@ -207,7 +208,10 @@ defmodule MLLP.Client do
   * `:use_backoff` - Specify if an exponential backoff should be used for connection. When an attempt
      to establish a connection fails, either post-init or at some point during the life span of the client,
      the backoff value will determine how often to retry a reconnection. Starst at 1 second and increases
-     it until reaching 180 seconds.  Defaults to `false`.
+     exponentially until reaching `backoff_max_seconds` seconds.  Defaults to `false`.
+
+  * `:backoff_max_seconds` - Specify the max limit of seconds the backoff reconection attempt should take,
+     defauls to 180 (3 mins).
 
   * `:reply_timeout` - Optionally specify a timeout value for receiving a response. Must be a positive integer or
      `:infinity`. Defaults to `:infinity`.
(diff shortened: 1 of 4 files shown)


test/client_test.exs
```diff
@@ -107,7 +107,7 @@ defmodule ClientTest do
       |> expect(:send, fn ^socket, ^packet -> :ok end)
       |> expect(:recv, fn ^socket, 0, :infinity -> {:ok, tcp_reply} end)
 
-      {:ok, client} = Client.start_link(address, port, tcp: MLLP.TCPMock)
+      {:ok, client} = Client.start_link(address, port, tcp: MLLP.TCPMock, use_backoff: true)
 
       expected_ack = %MLLP.Ack{acknowledgement_code: "AA", text_message: "You win!"}
```
starbelly, line 151: We should add some tests to ensure backoff is working as expected.

starbelly:
@pggalaviz The issue mentioned max attempts. I'm not sure that's needed anymore, so I would ignore. I was still coming up to speed with mllp myself when I opened the issue. The general idea is that an mllp connection from client to server is a connection that never terminates (when it can be helped). With that in mind, giving up after a certain number of attempts doesn't make sense here I believe.

lib/mllp/client.ex
```diff
@@ -193,24 +197,32 @@ defmodule MLLP.Client do
...
+  This function will raise a `ArgumentError` if an invalid `ip_address()` is provided.
 
   ## Options
 
   * `:auto_reconnect_interval` - Specify the interval between connection attempts. Specifically, if an attempt
      to establish a connection fails, either post-init or at some point during the life span of the client, the value
      of this option shall determine how often to retry a reconnection. Defaults to 1000 milliseconds.
 
-  * `:reply_timeout` - Optionally specify a timeout value for receiving a response. Must be a positive integer or 
+  * `:use_backoff` - Specify if an exponential backoff should be used for connection. When an attempt
```
starbelly, line 204: I think `use_backoff` be the default. What do you think @vikas15bhardwaj ?

vikas15bhardwaj: yes, we can make it default

lib/mllp/client.ex
```diff
@@ -458,11 +470,17 @@ defmodule MLLP.Client do
...
 
-  defp ensure_pending_reconnect_cancelled(state) do
+  defp ensure_pending_reconnect_cancelled(%{backoff: nil} = state) do
     :ok = Process.cancel_timer(state.pending_reconnect, info: false)
     %{state | pending_reconnect: nil}
   end
 
+  defp ensure_pending_reconnect_cancelled(%{backoff: backoff} = state) do
+    :ok = Process.cancel_timer(state.pending_reconnect, info: false)
+    {_, new_backoff} = :backoff.succeed(backoff)
```
vikas15bhardwaj, line 482: backoff reset may not be right here. `ensure_pending_reconnnect_cancelled/1` is also called from when stopping the connection before attempting a new connection. So its possible, connection was never successful but we still reset the backoff.

starbelly (approved):

PR #55, opened 2022-10-27
Do not fail on bad acks

nerdyworm:


nerdyworm:
sorry, wrong repo :)    In case your wondering, we are dealing with a system that doesn't fail well and these exceptions hide the AR/error message for us.

PR #63, opened 2023-04-10
Handle fragmented client responses

### LINKED ISSUE #62: Fragmented  Ack
We seem to have a problem where we send a message and receive the ack split into 2 TCP packets, in the first one is the MLLP start byte and the whole ML7 ack, in the second one is only the MLLP end bytes.

As a temporary solution, I have now adapted the MLLP library as follows:
```elixir
def handle_call({:send, message, options}, _from, state) do
    options1 = Map.merge(state.send_opts, options)
    telemetry(:sending, %{}, state)
    payload = MLLP.Envelope.wrap_message(message)

    case state.tcp.send(state.socket, payload) do
      :ok ->
        case state.tcp.recv(state.socket, 0, options1.reply_timeout) do
          {:ok, reply} ->
            reply =
              case state.tcp.recv(state.socket, 0, 1_000) do
                {:ok, reply2} -> reply <> reply2
                _ -> reply
              end

            telemetry(:received, %{response: reply}, state)
            {:reply, {:ok, reply}, state}
  ...
  end
```

How could we solve this more cleanly, preferably in such a way that we can also deal with more fragmented messages?

starbelly:
This commit resolves #62. Specifically, the number of chunks a response from a server will be pushed up the stack to the socket is non-deterministic, therefore we must loop until we've got response that is exactly an header, some data, and a trailer or a header, some data, trailer, and some trailing junk. 

Note that the latter should never happen with a sane MLLP server implementation, but we can't disregard the possibility either. With that said, it's best to return it all to the caller to let them deal with it vs making assumptions in regards to whether it's safe to truncate or not.


lib/mllp/client.ex
```diff
@@ -464,6 +468,34 @@ defmodule MLLP.Client do
...
+      {:ok, reply} ->
+        new_buf = buffer <> reply
+        check = byte_size(new_buf) - 3
+
+        case new_buf do
+          <<@header, _ack::binary-size(check), @trailer>> ->
+            {:ok, new_buf}
+
+          <<@header, _rest::binary>> ->
+            case :binary.split(new_buf, state.mllp_trailer, [:global]) do
```
starbelly: We use binary split here directly to avoid some overhead with say calling String.split/2. That is, calling `String.split/3`, which in turn will compute the scope for the split, and then finally call `:binary.split/3`. Instead, we just split with a global split which is equivlent to what String.split/3 would do for this case. 

What's more, we pass in pre-compiled binary pattern. In other words, latency matters here, especially per the request/response model of MLLP/HL7.

lib/mllp/client.ex
```diff
@@ -338,6 +342,7 @@ defmodule MLLP.Client do
       |> validate_options()
       |> maybe_set_default_options()
       |> put_socket_address()
+      |> Map.put(:mllp_trailer, :binary.compile_pattern(@trailer))
```
starbelly: Compile a binary pattern using the MLLP trailer and use this for fragmented response validation since we'll do this indefinitely.

lib/mllp/client.ex
```diff
@@ -464,6 +468,34 @@ defmodule MLLP.Client do
...
+          <<@header, _rest::binary>> ->
+            case :binary.split(new_buf, state.mllp_trailer, [:global]) do
+              [_, _ | _] ->
+                {:ok, new_buf}
+
+              _ ->
+                recv_ack(state, opts, new_buf)
+            end
+
+          _ ->
```
starbelly, line 506: Here's a rub, do we check that the buffer at least has the header? That is, it may be proceeded by junk, then VT, then some data, but still awaiting the rest.

IMO, we should not, but if that's the case then we should ditch the `:binary.split/3` as well, as then we're being strict, which is fine by me.

lib/mllp/client.ex
```diff
@@ -464,6 +468,34 @@ defmodule MLLP.Client do
     stop_connection(state, reason, "process terminated")
   end
 
+  defp recv_ack(state, opts, buffer) do
```
starbelly: There's one missing piece for this and that is a timeout or number of times we will attempt to pull chunks off the socket before returning an error.

starbelly: Note there are at least two ways to handle this. Modify the function `recv_ack` to decrement the timeout for each pass.  Other option is to spin up a task that calls this function, and recv merely uses a really long timeout.

The latter is unappealing because we might block the socket for way too long. 

This also leads to a problem we currently have, we allow for `:infinity` as the reply timeout, which is just generally problematic.  If infinity is allowed, we can block on the backend or on the socket, well for infinity.

starbelly: I pushed up an initial version of the former (timeout based loop)

starbelly: ~~Alas (and sorry for the noise), there may be a third option, that is specify the line delimiter (in this case it would be the trailer), I believe this can only be used an active mode, let me experiment...~~

test/client_test.exs
```diff
@@ -156,6 +158,96 @@ defmodule ClientTest do
...
+      |> expect(:recv, fn ^socket, 0, :infinity -> {:ok, ack_frag2} end)
+      |> expect(:recv, fn ^socket, 0, :infinity -> {:ok, ack_frag3} end)
+
+      assert(
+        {:ok, :application_accept, expected_ack} ==
+          Client.send(client, message)
+      )
+    end
+
+    test "when replies are fragmented and the last fragment is not received" do
```
starbelly, line 215: This is unfortunately a brittle test. If we keep this approach, then we might want to tag it as such and not run it by default.

lib/mllp/client.ex
vikas15bhardwaj: Curious, how much performance we are gaining here using `:binary.split` vs `String.split`?

lib/mllp/client.ex
```diff
@@ -464,6 +468,52 @@ defmodule MLLP.Client do
...
+      {:ok, reply} ->
+        new_buf = buffer <> reply
+        check = byte_size(new_buf) - 3
+
+        case new_buf do
+          <<@header, _ack::binary-size(check), @trailer>> ->
+            {:ok, new_buf}
+
+          <<@header, _rest::binary>> ->
+            case :binary.split(new_buf, state.mllp_trailer, [:global]) do
```
vikas15bhardwaj: we have 4 nested case here, can we split these into may be functions

lib/mllp/client.ex
vikas15bhardwaj: I don't think either, If message is fragmented, I don't think it will second fragmented part will have header. It should be continuation of first fragment.

lib/mllp/client.ex
starbelly: Right, that was indeed the plan, but wanted to keep the behaviour all in one place so the solution was clear.

lib/mllp/client.ex
starbelly: To be clear, you're in favor of being "strict"? That is, we don't concern ourselves with preceding or trailing junk. The message must be received according to the MLLP spec.

vikas15bhardwaj: Sorry about the confusion. I got confused with header pattern match and your question here.

For Ack to start, we must receive header and check it, right? If ACK is fragmented, rest of the fragment will just append to it until we get trailing char. Sorry maybe I am not understanding the question?  I don't think we need to be concerned about junk, it should still be send as per MLLP spec and if that's what you mean by strict, thats fine.

starbelly: Sounds good to me! :)

lib/mllp/client.ex
starbelly: Well, we'll be ditching this part, but it actually (at least on my machine) was not enough to justify the optimization (about a microsecond!).

lib/mllp/client.ex
```diff
@@ -464,6 +470,51 @@ defmodule MLLP.Client do
...
+
+  defp recv_ack(state, timeout, buffer) do
+    {res, elapsed} = do_recv(state, 0, timeout)
+
+    case res do
+      {:ok, reply} ->
+        new_buf = buffer <> reply
+        check = byte_size(new_buf) - 3
+
+        case new_buf do
```
starbelly, line 498: My general rule of thumb is one nested case expr? 👍 
More than one is on the path to shark fins. We could bust the nested case expression out into another function, but there's also value in keeping as much of the behaviour in on place (especially the call to recurse). Thoughts?

starbelly:
@vikas15bhardwaj  One other change we may want to consider is changing the default `reply_timeout` from `:infinity` to 60 or 120 seconds. I think that's definitely adequate, we will still accept `:infinity`, but I think 60 seconds as an example makes for a better debug experience.

lib/mllp/client.ex
```diff
@@ -464,6 +470,51 @@ defmodule MLLP.Client do
...
+    end
+  end
+
+  defp do_recv(state, length, :infinity) do
+    res = state.tcp.recv(state.socket, length, :infinity)
+    {res, nil}
+  end
+
+  defp do_recv(state, length, timeout) do
+    t1 = System.convert_time_unit(:erlang.monotonic_time(), :native, :microsecond)
```
starbelly: I think this conversion may be pointless... depends on how you look at it. With the conversion from and to milliseconds, we lose precision. That is to say, as long as some milliseconds were spent in the recv op, then we will decrement the timeout by 1ms. Given that, we may as well simply measure in millisecond and if the elapsed time is 0, then we simply decrement the timeout by 1. 

The exception would be if 0 microseconds were spent (i.e., we spent nanoseconds in the recv op) then we don't decrement (i.e., the way it currently is).

test/client_test.exs
```diff
@@ -156,6 +161,131 @@ defmodule ClientTest do
...
+        fn ^address, ^port, [:binary, {:packet, 0}, {:active, false}], 2000 ->
+          {:ok, socket}
+        end
+      )
+      |> expect(:send, fn ^socket, ^packet -> :ok end)
+      |> expect(:recv, fn ^socket, 0, 2 -> {:ok, ack_frag1} end)
+      |> expect(:recv, fn ^socket, 0, 1 -> {:ok, ack_frag2} end)
+
+      {:ok, client} =
+        Client.start_link(address, port, tcp: MLLP.TCPMock, use_backoff: true, reply_timeout: 2)
```
vikas15bhardwaj: On this test case, if we add another expect or may be stub of `:recv`, then it may not fail if we don't timeout before 3-4 calls to recv are made.

vikas15bhardwaj:
> @vikas15bhardwaj One other change we may want to consider is changing the default `reply_timeout` from `:infinity` to 60 or 120 seconds. I think that's definitely adequate, we will still accept `:infinity`, but I think 60 seconds as an example makes for a better debug experience.

We may break existing applications if they are relying on default infinity timeouts if we change this. I see your point, but applications today have option to change timeout if required anyway.

lib/mllp/client.ex
```diff
@@ -464,6 +470,74 @@ defmodule MLLP.Client do
...
+        check = byte_size(new_buf) - 3
+
+        case new_buf do
+          <<@header, _ack::binary-size(check), @trailer>> ->
+            {:ok, new_buf}
+
+          <<@header, _rest::binary>> ->
+            recv_ack(
+              state,
+              maybe_decrement_timeout(timeout, time_spent + elapsed),
```
starbelly: Had to adjust how we're doing this. If a client supplies a 5 millisecond timeout (as an example), and the ack/nack comes in 6 segments; in the previous implementation we would have bailed after about 20 microseconds. 

This version should be more accurate.

starbelly:
> We may break existing applications if they are relying on default infinity timeouts if we change this. I see your point, but applications today have option to change timeout if required anyway.

I'm not worried about that in that we are not 1.0, thus you should expect breaking changes, that said we could bump the minor segment of the version too.

test/client_test.exs
starbelly: I think I've resolved this, if I understand correctly, but let me know.

lib/mllp/client.ex
vikas15bhardwaj: I am wondering, if we should keep `reply_timeout` in `microsecond` for calculation, and only convert it before `:recv` call. Wouldn't this make your calculation more simple? 
`case recv_ack(state, options1.reply_timeout) do` call should be `case recv_ack(state, System.convert_time_unit(options1.reply_timeout), :millisecond, :microsecond)) do`

starbelly: I think I'm following you 🤔  I'm not sure which part it would simplify though. Maybe it's the ` timeout - div(spent, 1000)` ? 

If so, we'd have to do that anyway, the reason exemplified is : 

```
iex(1)> micros = System.convert_time_unit(5000, :millisecond, :microsecond)
5000000
iex(2)> micros - 1001
4998999
iex(3)> updated = micros - 1001
4998999
iex(4)> System.convert_time_unit(updated, :microsecond, :millisecond)
4998
```

So, we lose 2 ms vs 1, that big of decrement in the case of small reply_timeout values is a big difference I believe.

But I may be off here and you might be pointing at a simplification I'm not seeing or thinking of.

vikas15bhardwaj: This is the simplification, I was talking about. I don't seem to loose time in it. Every time we try to recv and get a response, we reduce our timeout in microsecond from elapsed time. if timeout becomes zero anytime, we timeout otherwise carry on.

```
  defp recv_ack(state, timeout) do
    recv_ack(state, timeout, <<>>)
  end

  defp recv_ack(_state, time_left, _buffer)
       when is_integer(time_left) and time_left <= 0 do
    {:error, :timeout}
  end

  defp recv_ack(state, time_left, buffer) do
    {res, elapsed} = do_recv(state, 0, time_left)

    case res do
      {:ok, reply} ->
        new_buf = buffer <> reply
        check = byte_size(new_buf) - 3

        case new_buf do
          <<@header, _ack::binary-size(check), @trailer>> ->
            {:ok, new_buf}

          <<@header, _rest::binary>> ->
            recv_ack(state, time_left - elapsed, new_buf)

          _ ->
            {:error, :invalid_reply}
        end

      {:error, _} = err ->
        err
    end
  end
```
```
iex(8)> micros = System.convert_time_unit(5000, :millisecond, :microsecond)
5000000
iex(9)> elapsed = 1000                                                     
1000
iex(10)> updated = micros - elapsed                                         
4999000
iex(11)> System.convert_time_unit(updated, :microsecond, :millisecond)      
4999
```

starbelly: RIght, it's when the elapsed is say 1001, or even 1 microsecond.  I did adjust to pass around microseconds though as it's easier to follow :)

vikas15bhardwaj: how about if we calculate elapsed time in `millisecond` itself. Worst case, we will over weight a bit but would never bail out early

vikas15bhardwaj: Based on the discussion, leaving this with microsecond to keep it more precise in timeout

vikas15bhardwaj (approved):

PR #64, opened 2023-04-12
Update default timeouts

starbelly:
- Use and set send_timeout by to 60 seconds in `socket_opts`
- change reply_timeout to 60 seconds
- change caller timeouts to infinity


lib/mllp/client.ex
```diff
@@ -635,11 +636,11 @@ defmodule MLLP.Client do
   @default_opts %{
     telemetry_module: MLLP.DefaultTelemetry,
     tls_opts: [],
-    socket_opts: []
+    socket_opts: [send_timeout: 60_000]
```
starbelly, line 660: send_timeout has two flavors via send_timeout_close, which defaults to true, a timeout in sending will result a disconnection of the conn. This can be disabled, but it's probably the best thing to do, especially with MLLP / HL7.

lib/mllp/client.ex
```diff
@@ -407,7 +411,11 @@ defmodule MLLP.Client do
               state
             )
 
-            new_state = maintain_reconnect_timer(state)
+            new_state =
```
starbelly, line 420: If we encounter an error while trying to recv it is most likely unrecoverable and thus the best course of action is hang up the phone to clear state between the client and the server. But we also need to provide users a way to opt out of this behavior (`close_on_recv_error`).

lib/mllp/client.ex
```diff
@@ -271,19 +275,19 @@ defmodule MLLP.Client do
...
   * `:reply_timeout` - Optionally specify a timeout value for receiving a response. Must be a positive integer or
-     `:infinity`. Defaults to `:infinity`.
+     `:infinity`. Defaults to 60 seconds.
   """
   @spec send(
           pid :: pid,
           payload :: HL7.Message.t() | String.t() | binary(),
           options :: ClientContract.send_options(),
-          timeout :: non_neg_integer()
+          timeout :: non_neg_integer() | :infinity
```
vikas15bhardwaj, line 290: We need to change it on `@callback` too?

lib/mllp/client.ex
vikas15bhardwaj: Is this understanding correct:
- MLLP.Client.send/4 timeout which defaults to `infinity` is the `gen_server` call timeout. It's not related to socket send_timeout.
- Socket send_timeout which now defaults to 60_000 can only be provided before starting a client connection.
- reply_timeout which defaults to 60_000 is for receiving ACK which can be provided on each send call using options `%{reply_timeout: 1_000}`

lib/mllp/client.ex
starbelly: Good catch :)

lib/mllp/client.ex
starbelly: That is 100% correct.

vikas15bhardwaj:
Do you think we should add these timeout details to either Client module or in README? I checked, we have reply_timeout in start_link doc, and generic socket_opts. May be we can add socket_opts send_timeout separately or some example.

starbelly:
> Do you think we should add these timeout details to either Client module or in README? I checked, we have reply_timeout in start_link doc, and generic socket_opts. May be we can add socket_opts send_timeout separately or some example.

In fact, I do, I think I wanted to see how you felt about the changes first before documenting them :) Also need to add another test. ❤️

lib/mllp/client.ex
```diff
@@ -228,6 +228,12 @@ defmodule MLLP.Client do
      Note that `:binary`, `:packet`, and `:active` can not be overridden. Default options are enumerated below.
       - send_timeout: Defaults to 60 seconds
 
+  * `:close_on_recv_error` - A boolean value which dictates whether the the client socket will be
```
vikas15bhardwaj: the the

vikas15bhardwaj (approved):

Issue #66, opened 2023-04-20
Consider re-working client to use acitve mode with gen_tcp 


starbelly:
Per #65 , #64 , and #63  we may want to consider re-working the client to use active mode provided by gen_tcp and leveraging gen_statem vs gen_server. 

Specifically, per the aforementioned PRs we're starting to essentially re-implement a tried and true feature of gen_tcp, namely `{active, true  |false |once | N }`.

starbelly:
Closed via #68

PR #68, opened 2023-07-10
Active socket, implementation with state machine

bokner:
Switch to use active mode for the socket. Addresses #66.


mix.exs
```diff
@@ -38,11 +38,14 @@ defmodule MLLP.MixProject do
...
       {:backoff, "~> 1.1.6"},
+      {:gen_state_machine, "~> 3.0"},
       {:ex_doc, "~> 0.24.2", only: :dev, runtime: false},
       {:dialyxir, "~> 1.1.0", only: [:dev, :test], runtime: false},
       {:mix_test_watch, "~> 1.0.2", only: :dev, runtime: false},
       {:mox, "~> 1.0.0", only: :test},
-      {:excoveralls, "~> 0.14.4", only: :test, runtime: false}
+      {:excoveralls, "~> 0.14.4", only: :test, runtime: false},
+      ## Debugging
+      {:replbug, github: "bokner/replbug"}
```
starbelly: We're going to need to make this optional. Especially because it pulls in from git repo.

mix.exs
```diff
@@ -38,11 +38,14 @@ defmodule MLLP.MixProject do
       {:ranch, "~> 1.8.0"},
       {:elixir_hl7, "~> 0.6.0"},
       {:backoff, "~> 1.1.6"},
+      {:gen_state_machine, "~> 3.0"},
```
starbelly: This IMHO should go. Erlang is generally not a great thing to wrap, especially this kind of wrapping IMO which is very much subject to rot and yet another dep to pull in. If it was in stdlib sure, but yeah.

test/client_test.exs
```diff
@@ -426,7 +299,7 @@ defmodule ClientTest do
         :connect,
         fn ^address,
            ^port,
-           [:binary, {:packet, 0}, {:active, false}, {:send_timeout, 60_000}],
+           [:binary, {:packet, 0}, {:active, true}, {:send_timeout, 60_000}],
```
starbelly: This makes perfect sense to me. Most use cases around this are not subject to copious amounts of requests, however we may want to think about it a little 🤔

starbelly (changes requested):

mix.exs
bokner: Yep, I removed it.

mix.exs
bokner: Done

test/client_test.exs
starbelly: Thought about this some... I think active once will be needed for some use cases, but we can defer to that until it is actually needed. It's pretty much add the complexity now or later :)

lib/mllp/client.ex
```diff
@@ -55,10 +55,10 @@ defmodule MLLP.Client do
   Upon successful start up via `start_link/4`, the  client will attempt to establish a connection to the given address
   on the provided port. If a connection can not be immediately established, the client will keep
   trying to establish a connection per the value of `:auto_reconnect_interval` which defaults to
-  1 second. Therefor it is possible that before a connection is fully established, the caller
+  1 second. Therefore it is possible that before a connection is fully established, the caller
```
starbelly, line 58: Love it!

lib/mllp/client.ex
```diff
@@ -125,12 +125,13 @@ defmodule MLLP.Client do
...
 
-  use GenServer
   require Logger
 
   alias MLLP.{Envelope, Ack, ClientContract, TCP, TLS}
 
-  @behaviour ClientContract
+  @behaviour MLLP.ClientContract
+
+  @behaviour :gen_statem
```
starbelly, line 134: ❤️ 🧡 💛 💚 💙 💜

lib/mllp/client.ex
bokner: Thanks, just wanted to show that I did read it :-)

lib/mllp/client.ex
```diff
@@ -356,205 +372,273 @@ defmodule MLLP.Client do
...
+  defp handle_closed(state) do
+    handle_error(:closed, state)
   end
 
+  ## Handle transport errors
+  defp handle_error(reason, state) do
+    Logger.error("Error: #{inspect(reason)}, state: #{inspect(state)}")
+
+    reply_to_caller({:error, new_error(get_context(state), reason)}, state)
+    |> stop_connection(reason, "closing connection to cleanup")
```
starbelly: Nit pick: 

```suggestion
    {:error, new_error(get_context(state), reason)}
    |> reply_to_caller(state)
    |> stop_connection(reason, "closing connection to cleanup")
```

lib/mllp/client.ex
```diff
@@ -356,205 +372,273 @@ defmodule MLLP.Client do
...
+
+  ## Handle the (fragmented) responses to `send` request from a caller
+
+  defp handle_received(_reply, %{caller: nil} = state) do
+    ## No caller, ignore
+    state
+  end
+
+  defp handle_received(reply, %{receive_buffer: buffer} = state) do
+    new_buf = (buffer && buffer <> reply) || reply
```
starbelly: Do you think we should switch to an iolist here? It's an optimization, but yeah.

josevalim: :+1: for io lists, even if it optimization doesn't matter, I'd say it is semantically better anyway. The buffer can always be initialized to an empty list as well, so this line becomes:

```suggestion
    new_buf = [buffer | reply]
```

PS: you could also initialize it to an empty string and always concatenate.

josevalim:
LGTM to me and also great reviews @starbelly! 💯

lib/mllp/client.ex
bokner: I agree with the suggestion to use an empty string instead of `nil`.
The problem with the suggested change though is that it breaks the code unless we rewrite the rest of `handle_received/2`.

starbelly: Indeed. Maybe commit the suggest or push the the rest? Or similar. I think that was implicit per the suggestion IMHO.

bokner: It can definitely be done, but I don't see why would we prefer dealing with io_lists vs. directly with binary. For one thing, the data will come as a binary, so I'm not sure why would we do:
```new_buf = [buffer | reply]``` (and then flatten the data?)
vs. 
```new_buf = buffer <> reply```

Comment (bokner) [reply lib/mllp/client.ex]: Having said this, the suggestion made me revisit the code, and I think [2c2c400](https://github.com/HCA-Healthcare/elixir-mllp/pull/68/commits/2c2c4004f4bcf2668eeea9dabd5991e30c9e8354) makes `handle_received/2` a bit more efficient by not explicitly checking the size of the buffer between header and trailer.

Comment (josevalim) [reply lib/mllp/client.ex]: `buffer <> reply` will copy both sides on every operation, given the buffer increases, the amount being copied is ever increasing. So it is more costly in terms of CPU and memory. If you do `[buffer | reply]` and at the end you do `IO.iodata_to_binary(buffer)`, then everything is just copied once (with no increasing costs).

Comment (starbelly) [reply lib/mllp/client.ex]: This may be a premature comment, but when it comes to checking if we have everything (i.e, a full ack) what we can do is something like : 

```
check = byte_size(new_fragment) - 2 
case {receiver_buffer, new_fragment]} do 
  {[^@header | _],  <<_::binary-size(check), @trailer>>]  -> Enum.join(receiver_buffer)
  _ -> :moar_data_coming
 end 
```

Something like that, this allows us to defer the concatenation until the very end 😄

Comment (starbelly) [reply lib/mllp/client.ex]: haha, I think we had a race condition on the suggestion here! 😄 Mine is flawed in that I suggested Enum.join, and IO.iodata_to_binary/1 is going to be much better here 😁

Comment (starbelly) [reply lib/mllp/client.ex]: Sorry for the noise, but there's one other check you have to make, it's an edge case. 

```
case {receiver_buffer, new_fragment]} do 
   {[^@header | _],  <<_::binary-size(check), @trailer>>]}  -> IO.iodata_to_binary(receiver_buffer)
    {[^@header | _],  <<_::binary-size(check - 1), @first_part_of_trailer::binary-size(1)>>, @carriage_return]}  -> IO.iodata_to_binary(receiver_buffer)
    _ -> :moar_data_coming
 end
```

There might be a nicer way to do that, but I'm not aware of one.

Comment (bokner) [reply lib/mllp/client.ex]: > `buffer <> reply` will copy both sides on every operation, given the buffer increases, the amount being copied is ever increasing. So it is more costly in terms of CPU and memory. If you do `[buffer | reply]` and at the end you do `IO.iodata_to_binary(buffer)`, then everything is just copied once (with no increasing costs).


Thanks, @josevalim
This makes sense. 
We will have to rewrite the code then to check every incoming chunk of data for the trailer being in there **_before_** we do [buffer | reply]. Which is what I am going to do :-)

File: lib/mllp/client.ex [hunk 8]
```diff
@@ -356,282 +373,359 @@ defmodule MLLP.Client do
...
+    end
   end
 
-  defp do_recv(state, length, timeout) do
-    timeout_in_ms = System.convert_time_unit(timeout, :microsecond, :millisecond)
-    t1 = System.monotonic_time(:microsecond)
-    res = state.tcp.recv(state.socket, length, timeout_in_ms)
-    t2 = System.monotonic_time(:microsecond)
-    {res, t2 - t1}
+  defp has_trailer?(packet) do
```
starbelly: Right, this doesn't catch the edge case. I think if you go with case expr and a pattern match it's much clearer what's going on, as I suggested previously. 

Edit: 

Something like this solves for the issue. 

```elixir
def foo(buffer, new_fragment) do 
  case {receiver_buffer, new_fragment]} do 
   {[^@header | _],  <<_::binary-size(check), @trailer>>]}  -> IO.iodata_to_binary([receiver_buffer | new_fragment])
    {[^@header | _],  <<_::binary-size(check - 1), @first_part_of_trailer::binary-size(1)>>, @carriage_return]}  -> IO.iodata_to_binary([receiver_buffer | new_fragment])
    _ -> :moar_data_coming
  end
end
```

bokner: If I understand correctly, the edge case you are referring to is data that follows the trailer? I will write the failing test and fix it. Also, we do not have to match on the header within the buffer except for when we get the first packet.

starbelly: Yes, specifically what can happen is the last byte is received on its own. There was a fix I put in MLLP.Reciever for this (there should also be a test case for it). 

Edit: 

To note, this is easy to miss because most protocols are sane (or saner) 😁

lib/mllp/client.ex
```diff
@@ -356,282 +377,363 @@ defmodule MLLP.Client do
...
+      false ->
+        Logger.debug("Client #{inspect(self())} received a MLLP fragment: #{reply}")
+        Map.put(data, :receive_buffer, new_buf)
     end
   end
 
-  defp do_recv(state, length, :infinity) do
-    res = state.tcp.recv(state.socket, length, :infinity)
-    {res, 0}
+  defp trailer_check(packet) do
```
starbelly: Interestingly, this handles an edge case, but not the one I was referring to :)  I haven't seen this one in production, but it could happen. 

Back to the original edge case first that I believe I wasn't clear about : 

The trailer consists of two bytes, the file separator byte and a byte for a carriage return. What can and does happen is you get everything in chunks (in various ways) all the way to the file separator, then the carriage return arrives on its own. 

As an example, you get the first chunk it contains the header, followed by some hl7 data, then maybe you get a second chunk it contains more hl7 data and the first byte of the trailer, and finally you get carriage return in its own "packet". 

As for the binary match approach, I'm not sure on that vs a straight pattern match. I would ask @josevalim for his opinion as well.

bokner: > Interestingly, this handles an edge case, but not the one I was referring to :) 

I think it does handle that case, too :-)
The code waits for a full trailer, and it won't matter if the carriage return comes separately.
The test case for this "last byte" case is a bit trickier to make, but I'll spend some time putting it together.
[The general test for chunked messages](https://github.com/HCA-Healthcare/elixir-mllp/blob/cb3ebb4e3335f296c5c3109fa22dca43233cca4e/test/client_test.exs#L193) is there though.

starbelly: RIght, this is where I still see a straight pattern match as suggested before as advantageous, there's nothing tricky going on. Especially if you put keep it local handle_received or recv_impl (we'll avoid copies).

bokner: I meant creating a unit test for that case is tricky, we'd need to do a mock that mimics that exact "last byte" case, or alternatively, expose the "parsing" logic as a public fun and write the test for it.

starbelly: A mock would be fine, since that's already in use. I'd also be quite ok with exposing the parsing logic (I've done precisely this in a previous mllp related project).

lib/mllp/client.ex
```diff
@@ -356,282 +381,383 @@ defmodule MLLP.Client do
...
+
+  ## The rest of MLLP (after the header was received)
+  defp handle_received(reply, data) do
+    receive_impl(reply, data)
+  end
+
+  def receive_impl(reply, %{receive_buffer: buffer, last_byte_received: last_byte} = data) do
+    new_buf = update_receive_buffer(buffer, reply)
+
+    case trailer_check(reply, last_byte) do
```
starbelly, line 617: Cool, we're now dealing with those two edges cases, but I'm not quite sure I understand why we don't just do all that here and (or in one of the handle_received clauses) with some simple pattern matching? Is there something you're seeing that I'm not?

I also realized we're appending vs pre-pending to the buffer, that shouldn't be a huge issue, in theory the acks should be small, but it is non-deterministic 🤔 

Edit: 

This would do the trick (and iirc how I've done it before), but still would like to know if you think there is a down side to this approach. 

This requires to pieces of state, header_received and eb_received. The nice thing IMHO is that all states are captured close together and in a single clause. We can also just prepend to the buffer. 

```elixir
defp handle_received(_reply, %{caller: nil} = data) do
    ## No caller, ignore
    data
  end

  defp handle_received(<<@header, _::binary>> = chunk, %{header_received: false} = data) do
    chunk_size = byte_size(chunk)
    full_trailer_check = chunk_size - 2
    eb_only_check = chunk_size - 1

    case chunk do
      <<_::binary-size(full_trailer_check), @trailer>> ->
        reply_to_caller({:ok, buffer_to_binary([chunk])}, data)

      <<_::binary-size(eb_only_check), @eb>> ->
        %{data | buffer: [chunk | data.buffer], header_received: true, eb_received: true}

      _ ->
        %{data | buffer: [chunk | data.buffer], header_received: true}
    end
  end

  defp handle_received(@trailer, %{header_received: true, eb_received: false} = data) do
    reply_to_caller({:ok, buffer_to_binary([@trailer | data.buffer])}, data)
  end

  defp handle_received(@eb, %{header_received: true, eb_received: false} = data) do
    %{data | buffer: [@eb | data.buffer], eb_received: true}
  end

  defp handle_received(chunk, %{header_received: true, eb_received: false} = data) do
    chunk_size = byte_size(chunk)
    trailer_check = chunk_size - 2
    eb_only_check = chunk_size - 1

    case chunk do
      <<_::binary-size(trailer_check), @trailer>> ->
        reply_to_caller({:ok, buffer_to_binary([chunk | data.buffer])}, data)

      <<_::binary-size(eb_only_check), @eb>> ->
        %{data | buffer: [chunk | data.buffer], eb_received: true}

      _ ->
        %{data | buffer: [chunk | data.buffer]}
    end
  end

  defp handle_received(@cr, %{eb_received: true} = data) do
    reply_to_caller({:ok, buffer_to_binary([@cr | data.buffer])}, data)
  end

  defp handle_received(chunk, data) do
    reply_to_caller({:error, :invalid_reply}, data)
  end

  defp buffer_to_binary(buffer) do
    buffer
    |> Enum.reverse()
    |> IO.iodata_to_binary()
  end
```

lib/mllp/client.ex
```diff
@@ -717,6 +847,10 @@ defmodule MLLP.Client do
     }
   end
 
+  defp get_context(%State{context: context}) do
```
starbelly, line 883: I think we can just pattern match in the fun clause to avoid the boolean logic.

lib/mllp/client.ex
```diff
@@ -356,282 +381,383 @@ defmodule MLLP.Client do
...
+
+      b ->
+        [b | packet]
     end
   end
 
-  defp do_recv(state, length, :infinity) do
-    res = state.tcp.recv(state.socket, length, :infinity)
-    {res, 0}
+  defp buffer_to_binary(buffer) when is_list(buffer) do
```
starbelly: I don't think we need this function since the buffer will always be a list.

lib/mllp/client.ex
bokner: Not sure, it doesn't look much clearer to me than the current implementation. Also, I do not think that pattern matching on the chunk size several times  is much better than using `binary:match/2`.

The current code detects if there is no header in the very first packet, so it doesn't have to do more pattern matching. That makes `header_received=false` to be the same as `receive_buffer = []`, hence we probably don't need an additional field here. 

Wrt prepending: I think [buffer | reply] is a cons-cell operation, so we are not appending here. So it should be O(1) regardless of the ordering. Could of course be wrong here.

starbelly: > Not sure, it doesn't look much clearer to me than the current implementation. Also, I do not think that pattern matching on the chunk size several times is much better than using `binary:match/2`.

Ok cool, I wasn't sure if there was some egregious or something nasty lurking in my suggestion, but it sounds like it's a matter of taste. In that regard, I don't have terribly strong feelings on this one.

> The current code detects if there is no header in the very first packet, so it doesn't have to do more pattern matching. That makes `header_received=false` to be the same as `receive_buffer = []`, hence we probably don't need an additional field here.

This is a great point, you have a bail early clause, mine has to bail late. 



> Wrt prepending: I think [buffer | reply] is a cons-cell operation, so we are not appending here. So it should be O(1) regardless of the ordering. Could of course be wrong here.

Nope! You are 100% right 😄

lib/mllp/client.ex
```diff
@@ -356,282 +381,384 @@ defmodule MLLP.Client do
...
+  defp trailer_check(packet, _last_buffer_byte) do
+    case :binary.match(packet, @trailer) do
+      :nomatch -> false
+      {pos, @trailer_length} when pos + @trailer_length == byte_size(packet) -> true
+      _ -> :data_after_trailer
+    end
   end
 
-  defp update_recv_time_credit(:infinity, _), do: {:infinity, 0}
+  defp update_receive_buffer(buffer, packet) do
```
starbelly, line 650: This is actually beautiful in that you're creating an improper list, which is an optimization ❤️

mfos239:
Confirmed the latest changes fix this issue https://github.com/HCA-Healthcare/elixir-mllp/pull/65#issuecomment-1636466103
Everything seems to be working great on this end. Thanks so much @bokner and @starbelly!

lib/mllp/client.ex
```diff
@@ -356,282 +381,384 @@ defmodule MLLP.Client do
...
       )
+    end)
+  end
 
-      state.tcp.close(state.socket)
-    end
+  defp reply_cleanup(%State{} = data) do
+    data
+    |> Map.put(:caller, nil)
+    |> Map.put(:receive_buffer, [])
```
starbelly, line 690:
```suggestion
    data
    |> Map.put(:caller, nil)
    |> Map.put(:receive_buffer, [])
    |> Map.put(:last_byte_received, nil)
```

lib/mllp/client.ex
```diff
@@ -356,282 +381,384 @@ defmodule MLLP.Client do
...
-    {res, 0}
+  ## "Split trailer" case
+  ##
+
+  defp trailer_check(<<@cr, rest::binary>>, @eb) do
+    if rest == "" do
+      true
+    else
+      :data_after_trailer
+    end
```
starbelly:
```suggestion
  defp trailer_check(@cr, @eb), do: true

  defp trailer_check(<<@cr,  _::binary>>, @eb), do: :data_after_trailer
```

lib/mllp/client.ex
```diff
@@ -356,282 +381,384 @@ defmodule MLLP.Client do
...
+    case :binary.match(packet, @trailer) do
+      :nomatch -> false
+      {pos, @trailer_length} when pos + @trailer_length == byte_size(packet) -> true
+      _ -> :data_after_trailer
+    end
   end
 
-  defp update_recv_time_credit(:infinity, _), do: {:infinity, 0}
+  defp update_receive_buffer(buffer, packet) do
+    case buffer do
```
starbelly: We don't need to case here. It should be fine to always `[buffer | packet]`

lib/mllp/client.ex
```diff
@@ -356,282 +381,384 @@ defmodule MLLP.Client do
...
-    time_charged = div(time_spent, 1000) * 1000
-    time_owed = time_spent - time_charged
-    {time_left - time_charged, time_owed}
+      b ->
+        [b | packet]
+    end
+  end
+
+  defp reply_to_caller(reply, %{caller: caller, context: context} = data) do
+    caller && :gen_statem.reply(caller, format_reply(reply, context))
```
starbelly, line 655: I think we can avoid the boolean logic here, in fact if we're ending up here when caller is `nil`, we have a bug somewhere further up and it would be best to crash.

lib/mllp/client.ex
bokner: The reason for having `buffer = []` case is to avoid creating a nested list with []. That is:
```elixir
iex(14)> buffer = []
[]       
iex(15)> reply = <<"abc">>
"abc"    
iex(16)> [buffer | reply] 
[[] | "abc"]
```

josevalim: Creating such a list should be fine :)

lib/mllp/client.ex
bokner: This is to avoid situations where the receiver sends more packets after full MLLP has been received.
It would not be a bug on our side, but rather a receiver misbehaving. Do you think we should crash in this case?

lib/mllp/client.ex
bokner: Sure :-) but why do it if we can avoid it?
Probably a nitpick on my side :-)

josevalim: To avoid unecessary conditionals in the code. :)

starbelly: Right, so it probably comes down to whether it bugs you to see it if you're debugging, otherwise, they should cost the same amount of memory in words. 

```elixir
ex(1)> buffer = []
[]
iex(2)> buffer = [buffer | "a"]
[[] | "a"]
iex(3)> buffer = [buffer | "b"]
[[[] | "a"] | "b"]
iex(4)> :erts_debug.flat_size(buffer)
16
iex(5)> :erts_debug.size(buffer)
16
iex(6)> buffer1 = ["a"]
["a"]
iex(7)> buffer1 = [buffer1 | "b"]
[["a"] | "b"]
iex(8)> :erts_debug.flat_size(buffer1)
16
iex(9)> :erts_debug.flat_size(buffer)
16
```

No strong feelings here, just noticed 😄

lib/mllp/client.ex
starbelly: Seems like we'd hit the clause at line 577, no? This raises an interesting question none the less. Crashing aside, one place a `RST` is appropriate is when the the remote sends data that is unexpected. I wonder if we should shutdown in this case? Feels heavy handed, but yeah 🤔

lib/mllp/client.ex
bokner: I would think that flattening nested lists into a binary would work better for lesser nested ones :-)
But yeah, let's go with less code instead. No strong feelings here as well :-)

lib/mllp/client.ex
bokner: > Seems like we'd hit the clause at line 577, no? 

You are correct, we are probably covered for the "more packets" case. Although I'm not entirely sure, as there could be a race condition with the timer in the "receiving" state. There is actually a more serious reason for checking if we have a caller - the receiver can disconnect at any given time. If it happens while the 'send' call is in progress, we notify the caller, otherwise, we don't (no caller there).

If we want to shut down on the "more packets", the natural place is around line 577 indeed.

starbelly: > Although I'm not entirely sure, as there could be a race condition with the timer in the "receiving" state.

Not sure, will have to stare a bit more and we'll both have to attempt to put it through the ringer :) 

I think shutting down makes sense, I think we can just use `maybe_close` , we might need to make a modification though (i.e., maybe sometimes we want to do a graceful close and maybe some times an abortive close).

 Right now maybe_close will attempt to do a graceful close, in this case it's not what we want. We want `:gen_tcp.shutdown(socket, :read_write)`. This is the same behavior as `send_timeout_close: true`, which ends up resulting in a `sock_select()` call that does an abortive close (`RST`).  This should be double checked though. 

Speaking of `send_timeout_close: true`, if we don't have it this set as the default in sock opts we should, the only reason it's false in otp is for backwards compat reasons (though it should be configurable by the user).

starbelly: the send_timout_close bit was resolved by your last commit @bokner. WE still need to hang up the phone on `receive_timeout` as well. Basically, per the nature of MLLP and the implicit behaviour when shipping HL7 around, there's may be no recourse in a situation where the receive timed out. 

Specifically, the client can either hang up the phone or make assumptions about how the server is implemented. In this regard, both the client and server may end up in a bad state and form an error loop. One example where a receive timeout can result in  bad things happening is you peel off part of the response, you time out waiting for the rest. The caller of the client decides to send again, yet there still data in transit (i.e., not in the local buffers and bubbled up to our app yet), the send happens, and now the rest of the ack from a previous response arrives. Meanwhile, the remote might get what you just sent and either accept it, and send back a nack or hang up the phone. 

Tricky business 😁 

Thoughts?

bokner: The downside of dropping the connection on the client side for whatever reason (say, `receive_timeout`) is that then it would have to reconnect again. Maybe it is a better recourse after all.

I guess the sane client would try to check for connection on any error anyway.

starbelly: Yeah, and to add to that, there needs to be options for all of it, that really bothers me, I think having too many options is never great, but sadly since MLLP isn't a much of a protocol, the behaviour is going to differ from system to system. All we can do is try to provide a mostly sane set of defaults that cover most cases. If I go by the "protocol" and how tcp works, it leads me here to these decisions.

starbelly: I tell ya what, let's not agonize over this right now. We're not at 1.0, code is easy to change, etc. Let's get test passing, merge, and take it from there.

starbelly (approved):
Beautiful! ❤️🧡💛💚💙💜

PR #71, opened 2023-07-19
Backoff reset

bokner:
- Reset the backoff timer on failing `send`, so the client waits less for the next `reconnect` attempt.
- Set `keepalive=true` as the default socket option.


starbelly (approved):
Amazing catches! 😄

PR #73, opened 2023-08-17
Custom SSL transport option

bokner:
This change allows to use a custom implementation of SSL transport. The default option is `:ranch_ssl`. 
The implementation module must implement the `ranch_transport` behavior. 
[The example of implementation](https://github.com/bokner/elixir-mllp/blob/ssl_transport_opt/lib/mllp/tls_logging_transport.ex) (which logs peer info for unsuccessful TLS handshakes) is included.


starbelly (approved):
❤️🧡💛💚💙💜

PR #76, opened 2023-09-29
Add reference to options type

### LINKED ISSUE #75: Dialyzer not happy when `ref:` passed in receiver child_spec
I'm new to this library, but noticed that Elixir-LS was not happy with our invocation of `MLLP.Receiver.child_spec/1` - we pass a reference, but the options type does not allow for that. 

Unless I'm missing something, this should be a simple matter of allowing `ref` in `options()`?

cdegroot:
This makes the options() type match what's allowed to pass in through [child_spec/1](https://eng.d2iq.com/blog/a-long-journey-to-cross-platform-developer-tooling-utopia-for-now/) as well.

Fixes https://github.com/HCA-Healthcare/elixir-mllp/issues/75

### DIFF
--- a/lib/mllp/receiver.ex
+++ b/lib/mllp/receiver.ex
@@ -41,7 +41,8 @@ defmodule MLLP.Receiver do
           dispatcher: module(),
           packet_framer: module(),
           transport_opts: :ranch.opts(),
-          context: map()
+          context: map(),
+          ref: reference()
         ]
 
   @behaviour :ranch_protocol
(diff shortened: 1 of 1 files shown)


lib/mllp/receiver.ex
```diff
@@ -41,7 +41,8 @@ defmodule MLLP.Receiver do
           dispatcher: module(),
           packet_framer: module(),
           transport_opts: :ranch.opts(),
-          context: map()
+          context: map(),
+          ref: reference()
```
starbelly: This should be `term()` here since we pass it on to ranch's child spec, which takes a ref argument of any. It's a bit confusing because in the code we pass this in via `args.receiver_id`, which defaults to `make_ref()` if no `ref` arg is given.

starbelly:
```suggestion
          ref: term()
```

starbelly:
Thank you!!! ❤️

PR #81, opened 2024-06-21
Bump HL7 to 0.9.1

mukulmajmudar:
- Update to elixir_hl7 0.9.1
- Bump our version to 0.9.5


starbelly:
Hi, appologies a response on this has taken so long, I'm still mulling over whether to update this yet or not. It will eventually be done, but hl7 is in a lot of flux and need to ensure this update won't effect existing use cases in a negative way. Thanks for opening!

PR #84, opened 2024-11-04
Delegate to :ssl for setopts/2 when transport is tls

starbelly:
Previously we always called `:inet.setopts/2` when a timeout was detected to set `{linger, {true, 0}}` to immediately tear down the socket before closing the connection. In doing so and in the case of tls the sslsocket structure gets passed to `:prim_inet.setopts/2` where the first argument is expected to be a port identifier.


test/client_and_receiver_integration_test.exs
```diff
@@ -580,7 +580,7 @@ defmodule ClientAndReceiverIntegrationTest do
     @tag client_cert: "tls/server/server_certificate.pem"
     @tag keyfile: "tls/server/private_key.pem"
 
-    @tag reason: [:handshake_failure]
+    @tag reason: [:unsupported_certificate]
```
vikas15bhardwaj, line 583: this test is failing on my local. I also reverted this change. It does not return an error

vikas15bhardwaj:
should we bump version?

test/client_and_receiver_integration_test.exs
vikas15bhardwaj: I think erlang version should be `26.2.5.6`, test worked after I changed to this version

vikas15bhardwaj:
In WP we are using elixir `1.15.7` version, won't we need to update WP elixir version before using this version of `mllp`?

test/client_and_receiver_integration_test.exs
starbelly: it should be 😄

starbelly:
> In WP we are using elixir `1.15.7` version, won't we need to update WP elixir version before using this version of `mllp`?

That won't be needed.

starbelly:
> should we bump version?

We can do that.

vikas15bhardwaj:
> > In WP we are using elixir `1.15.7` version, won't we need to update WP elixir version before using this version of `mllp`?
> 
> That won't be needed.

how does this work? will `mllp` be compiled to `1.15.7` version within WP?

PR #86, opened 2025-01-21
Reconnect on send error option.

starbelly:
Given the number of components and complex set of states for a tcp connection, from erlang ssl to sockets at the OS, it's always possible to end up in a bad state and it's not always possible to safe guard against that, especially when new behavior or bugs are introduced in erlang/otp. As such it makes sense to offer
an option to do a hard reset if we get a transport error when trying to send to mitigate such effects to put the back into a healthy state.

This commit introduces the reconnect_on_send_error option for just that purpose, and defaults to true.


lib/mllp/client.ex
```diff
@@ -482,7 +490,21 @@ defmodule MLLP.Client do
         )
 
         error_reply = {:error, new_error(:sending, reason)}
-        {:keep_state_and_data, [{:reply, from, error_reply}]}
+
+        case data.reconnect_on_send_error do
+          true ->
+            data = stop_connection(data, reason)
+
```
vikas15bhardwaj: should we log a notice message here, it maybe good to have some visibility.

starbelly: Agreed, and done.

vikas15bhardwaj (approved):

PR #90, opened 2025-11-15
Add receiver telemetry support

starbelly:


vikas15bhardwaj:
Are you think apps such as waterpark will use telemetry event to find out if connection started or stopped. It feels weird to do so.

starbelly:
> Are you think apps such as waterpark will use telemetry event to find out if connection started or stopped. It feels weird to do so.

tbh, yes. If I happened along, I think I would be sad if this didn't emit telemetry. 

That aside, it provides a nice way to provide that event information without adding more options (i.e., callback for connect, disconnect, etc.) with the added fact that telemetry is super efficient.

vikas15bhardwaj:
Do you think adding couple of callbacks would be too much of work to do? Adding telemetry is fine, but I don't think we should use them for custom logic.

starbelly:
> > Are you think apps such as waterpark will use telemetry event to find out if connection started or stopped. It feels weird to do so.
> 
> tbh, yes. If I happened along, I think I would be sad if this didn't emit telemetry.
> 
> That aside, it provides a nice way to provide that event information without adding more options (i.e., callback for connect, disconnect, etc.) with the added fact that telemetry is super efficient.

I think that's point though. There's nothing custom in here 😄 To a user of a lib it's just generic events (when the connection  is established, when it terminates and why, and how long the client was connected for). I would expect this from such a library these days.

It's just waterpark can be make use of the events do some things it needs to do.

lib/mllp/receiver.ex
```diff
@@ -443,4 +480,34 @@ defmodule MLLP.Receiver do
...
+    :ok
+  end
+
+  defp telemetry(event_name, measurements, meta, %{telemetry_module: telemetry_module} = state) do
+    measurements = add_measurements(List.last(event_name), measurements, state)
+
+    telemetry_module.execute(
+      [:mllp, :receiver] ++ event_name,
+      add_measurements(List.last(event_name), measurements, state),
+      Map.merge(meta, state.framing_context.receiver_context)
```
vikas15bhardwaj: Is there a risk of adding too much metadata? or from lib perspective, it does not matter

starbelly: No, the flow of telemetry is basically, it's all just function calls. The only thing that gets put in ets is the handler attached. In other words, there's no data held on to to in ets or passed in messages. It's up the attached handler what to do with the data, if anything. 

I have pushed up an event_callback option so we can see what that looks like side by side. We could have both (i.e., telemetry does not emit receiver context in meta, but you can supply an event callback fun to get that data). 

If we agree, then will clean up.

vikas15bhardwaj (approved):

PR #92, opened 2026-03-27
Add response handler for MLLP.Client.send/xx

bokner:
### The issue

The return of MLLP.Client.send/2,3,4 is insufficient for handling ACKs.

- In case the outgoing message is a raw HL7, the return does not contain ACK info:
```elixir
iex(34)> raw = HL7.Examples.wikipedia_sample_hl7
..................
iex(35)> Client.send(pid, raw)
{:ok,
 "MSH|^~\\&|SuperOE|XYZImgCtr|MegaReg|XYZHospC|20060529090131-0500||ACK^A01^ACK|01052901|P|2.5\rMSA|AA|01052901|A real MLLP message dispatcher was not provided\r"}
```

- In case the outgoing message is an `HL7.Message` struct, the return does not contain the ACK message.

```elixir
iex(38)> Client.send(pid, HL7.Message.new(raw))

{:ok, :application_accept,
 %MLLP.Ack{
   acknowledgement_code: "AA",
   text_message: "A real MLLP message dispatcher was not provided",
   hl7_ack_message: nil
 }} 
```

### The solution

I think that the best solution would be to unify the output of MLLP.Client.send/xx

This, however, would break an existing API.

The alternative (as per this PR) is to provide an option for the callback function that exposes a "wrapped" raw response.
This is implemented as a side effect (similar to the telemetry handler), so the API would be preserved.

### DIFF
--- a/lib/mllp/client.ex
+++ b/lib/mllp/client.ex
@@ -651,7 +651,14 @@ defmodule MLLP.Client do
     receive_impl(reply, data)
   end
 
-  def receive_impl(reply, %{receive_buffer: buffer, last_byte_received: last_byte, response_handler: response_handler} = data) do
+  def receive_impl(
+        reply,
+        %{
+          receive_buffer: buffer,
+          last_byte_received: last_byte,
+          response_handler: response_handler
+        } = data
+      ) do
     new_buf = update_receive_buffer(buffer, reply)
 
     case trailer_check(reply, last_byte) do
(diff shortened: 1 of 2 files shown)


bokner:
> I think that the best solution would be to unify the output of MLLP.Client.send/xx

More concretely, have a return for raw HL7 input to have the same structure as for "HL7 struct" input.
Also, it appears that `hl7_ack_message` is never populated.

bokner:
Closing. #93 would be sufficient to be able to process in-band response from `MLLP.Client.send/xx`.

